PR #8, opened 2018-11-28
add vendor directory

aanm:
Signed-off-by: André Martins <andre@cilium.io>


jrajahalme:
@aanm The generated Go files in`go/` are not consumed in this repo, so I think we don't need `vendor/` directory here?

PR #11, opened 2019-01-31
allow listener reuse

jrajahalme:
Make it possible to share listeners between many endpoints by:
- using source or destination IP address (for egress and ingress, respectively) as the policy name, unless explicitly configured
- detect the proxylib protocol (if any) from the applicable policy


WORKSPACE
```diff
@@ -7,8 +7,8 @@ workspace(name = "cilium")
 #
 # No other line in this file may have ENVOY_SHA followed by an equals sign!
 #
-ENVOY_SHA = "4ef8562b2194f222ce8a3d733fb04c629eaf0667"
-ENVOY_SHA256 = "bbb09ff2048bb2b14d8fb9ec957c6ddc7bfee147cca8d51f3d42b4e8084860bf"
+ENVOY_SHA = "4a9ff62ac703117cbacb0db1208979e755995762"
```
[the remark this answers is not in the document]
jrajahalme: @mrostecki We'll do a separate PR for Envoy update for v1.9.1 once it is released.

cilium/conntrack.cc
```diff
@@ -0,0 +1,282 @@
...
+      ct_maps6_.erase(ct6);
+      ENVOY_LOG(trace, "cilium.bpf_metadata: Closed IPv6 conntrack map {}", name);
+    }
+  }
+}
+
+CtMap::CtMap(const std::string &bpf_root) : bpf_root_(bpf_root) {
+  // Try to open the bpf maps from Cilium specific paths
+  openMap4("global");
+  openMap6("global");
```
joestringer: Should we try to error-check here / catch&log failures early? Presumably the CT map should be guaranteed to exist when the proxy is launched?

cilium/conntrack.cc
```diff
@@ -0,0 +1,282 @@
...
+  }
+
+  if (ip->version() == Network::Address::IpVersion::v4) {
+    // Lock for the duration of the map lookup and conntrack lookup
+    std::lock_guard<std::mutex> guard(maps_mutex_);
+    auto it = ct_maps4_.find(name);
+    if (it == ct_maps4_.end()) {
+      it = openMap4(name);
+    }
+    if (it == ct_maps4_.end()) {
```
joestringer: nit: Seems like this if statement is the exact same check as three lines up, doesn't need to be a separate block.

cilium/conntrack.cc
```diff
@@ -0,0 +1,282 @@
...
+      ENVOY_LOG(error, "cilium.bpf_metadata: No IPv4 conntrack map {}", name);
+      return 0;
+    }
+    auto ct = it->second.get();
+    if (!ct->ctmap4_tcp_.lookup(&key4, &value)) {
+      ct_maps4_.erase(it); // flush the map to force reload after each failure.
```
joestringer: nit: I assume here the `flush` is actually meaning to remove the specific CT map from the global maps index, not actually flushing a CT table?

cilium/conntrack.cc
joestringer: I guess we don't actually know if this is intentional, maybe the user just configured Cilium with IPv4/IPv6 disabled... in which case warning about it wouldn't really be right---info, if anything.

cilium/conntrack.cc
```diff
@@ -0,0 +1,278 @@
...
+      return 0;
+    }
+  } else { 
+    // Lock for the duration of the map lookup and conntrack lookup
+    std::lock_guard<std::mutex> guard(maps_mutex_);
+    auto it = ct_maps6_.find(map_name);
+    if (it == ct_maps6_.end()) {
+      it = openMap6(map_name);
+    }
+    if (it == ct_maps6_.end()) {
```
mrostecki: Here the statement is also the same as in line 244. Those blocks can be merged.

cilium/conntrack.cc
mrostecki: +1

If ipv4 or ipv6 are disabled, then specific maps don't even exist.

PR #15, opened 2019-12-06
Fix proxy compilation failure on arm64

Jianlin-lv:
'__aligned_u64' redefined causes the compilation fail
with the following error:
"./linux/type_mapper.h:16: error: "__aligned_u64" redefined [-Werror]"

Signed-off-by: Jianlin Lv <Jianlin.Lv@arm.com>


jrajahalme:
@Jianlin-lv Thanks for the fix!

PR #16, opened 2020-02-19
Pr/jrajahalme/envoy update 1.13.0

jrajahalme:


Makefile
```diff
@@ -92,7 +92,6 @@ $(CILIUM_ENVOY_BIN) $(CILIUM_ENVOY_RELEASE_BIN): force
 	@$(ECHO_BAZEL)
```
aanm: For this particular commit. Do we need to ship `cilium-envoy-unstripped` in the envoy docker image? This will increase the image size (by how much?), is it really necessary? why can't we create 2 docker images, one for debug and the other one containing the stripped binary only? I can help you with this if you want. I assume we only need `cilium-envoy-unstripped` for debug purposes so we can request for the backtrace and decode envoy stack traces locally.

Dockerfile.builder
```diff
@@ -6,7 +6,7 @@
 #
 # Using cilium-builder as the base to ensure libc etc. are in sync.
 #
-FROM quay.io/cilium/cilium-builder:2019-07-11 as builder
+FROM quay.io/cilium/cilium-builder:2020-02-13 as builder
```
aanm: There's a more recent one `2020-02-19`

Makefile
jrajahalme: @aanm Help would be appreciated. To be sure the unstripped version is exactly the same build, we must only build once. But then we need to modify the Git repo for the "stripped build" to refer to the unstripped build, by the time which the SHA of the repo has now changed, and is not reflecting the build SHA anymore. I guess we could just tag the stripped image with the actual build SHA instead of the GIT SHA, but I'm open to other suggestions as well!

aanm: @jrajahalme my suggestion will be to create a new repo, `quay.io/cilium/cilium-envoy-binaries`, that repository will need a new Dockerfile that creates envoy stripped and unstripped. 

Then, the ./Dockerfile on the root of this project would be something like:

```dockerfile
FROM quay.io/cilium/cilium-envoy-binaries:<commit-ID> as builder

FROM scratch
LABEL maintainer="maintainer@cilium.io"
COPY --from=builder /tmp/install/cilium-envoy-stripped /cilium-envoy
```
This one would be to create the image `quay.io/cilium/cilium-envoy`

And probably would be better to have another Dockerfile for the unstripped version:
```dockerfile
FROM quay.io/cilium/cilium-envoy-binaries:<commit-ID> as builder

FROM scratch
LABEL maintainer="maintainer@cilium.io"
COPY --from=builder /tmp/install/cilium-envoy-unstripped /cilium-envoy
```

This one would be to create the image `quay.io/cilium/cilium-envoy-unstripped`

Makefile
```diff
@@ -119,11 +118,8 @@ $(CHECK_FORMAT): force-non-root
 
 install: force-root
 	$(INSTALL) -m 0755 -d $(DESTDIR)$(BINDIR)
-	$(INSTALL) -m 0755 -T $(CILIUM_ENVOY_BIN) $(DESTDIR)$(BINDIR)/cilium-envoy
-# Strip only non-debug builds
-ifeq "$(findstring -dbg,$(realpath bazel-bin))" ""
-	$(STRIP) $(DESTDIR)$(BINDIR)/cilium-envoy
-endif
+	$(INSTALL) -m 0755 -T $(CILIUM_ENVOY_BIN) $(DESTDIR)$(BINDIR)/cilium-envoy-unstripped
+	$(STRIP) -o $(DESTDIR)$(BINDIR)/cilium-envoy $(DESTDIR)$(BINDIR)/cilium-envoy-unstripped
```
tklauser: Won't this strip the wrong binary (`cilium-envoy-unstripped`, but `cilium-envoy should be stripped)?

jrajahalme:
@aanm I left the unstripped binary out of this for now.

PR #17, opened 2020-02-24
Add Arm64 Support for Envoy image building

TrevorTaoARM:
Add Arm64 Dockerfile.builder and Dockerfile support for
image building on its builder and envoy image itself.

At the same time, we add the multi-arch support with
fat-manifest for both aarch64 and x86_64. After pushing
the images for the 2 arches, we can just use the cmd:
make docker-envoy-manifest
to enable multi-arch of both images which would greatly
facilitate the image pulling and use.

This patch would not affect the current use and building
for those of x86_64 platform with the following consideration:
1. The current 'amd64' arched images would be tagged as the
default images as the original;
2. The current building images tagged with 'SOURCE_VERSION'
would be tagged as the "latest", which would facilitate its use and
the local building of following images.

Signed-off-by: trevor tao <trevor.tao@arm.com>

### DIFF
--- a/Dockerfile.arm64
+++ b/Dockerfile.arm64
@@ -1,0 +1,26 @@
+#
+# Cilium incremental build. Should be fast given builder-deps is up-to-date!
+#
+# cilium-builder tag is the Git SHA of the compatible build image commit.
+# Keeping the old images available will allow older versions to be built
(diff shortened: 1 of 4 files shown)


TrevorTaoARM:
Abandon this PR with a new PR:
https://github.com/cilium/proxy/pull/20

PR #18, opened 2020-02-24
Add aarch64 support for Bazel install

TrevorTaoARM:
Add aarch64 support for tools/install_bazel.sh
which would install aarch64 release of Bazel when
meets arm64 platform.

Signed-off-by: trevor tao <trevor.tao@arm.com>

PR #20, opened 2020-03-18
Add Arm64 Support for Envoy image building

TrevorTaoARM:
Add Arm64 Support for Envoy image building
    
Add Arm64 Dockerfile.builder and Dockerfile support for
image building on its builder and envoy image itself.
    
Uses multi-arch enabled Dockerfile.builder and Dockerfile to
enable image building on both amd64 and arm64 platform.
    
The multi-arch support with fat-manifest for both arm64 and amd64
is also added, i.e.,after pushing the images for the 2 arches,
we can just use the cmds:
        make envoy-builder-manifest
        make docker-envoy-manifest
to enable multi-arch of both images which would greatly
facilitate the image pulling and use.

This patch would not affect the current use and building
for those of x86_64 platform with the following consideration:
1. The current 'amd64' arched images would be tagged as the
default images as the original;
2. The current building images tagged with 'SOURCE_VERSION'
would be tagged as the "latest", which would facilitate its use and
the local building of following images.

Signed-off-by: trevor tao <trevor.tao@arm.com>

### DIFF
--- a/Dockerfile.builder.multi_arch
+++ b/Dockerfile.builder.multi_arch
@@ -1,0 +1,66 @@
+#
+# Builder dependencies. This takes a long time to build from scratch!
+# Also note that if build fails due to C++ internal error or similar,
+# it is possible that the image build needs more RAM than available by
+# default on non-Linux docker installs.
+#
+# Using cilium-builder as the base to ensure libc etc. are in sync.
+#
+# The base image here should be multi-arched beforehand when used for non-amd64
+# platforms, e.g, arm64 platform.
+#
+FROM quay.io/cilium/cilium-builder:2020-05-05 as builder
+
+LABEL maintainer="maintainer@cilium.io"
+
+ARG ARCH=amd64
+
+WORKDIR /go/src/github.com/cilium/cilium/envoy
+COPY . ./
+
+#
+# Additional Envoy Build dependencies
+#
+RUN apt-get update \
+    && DEBIAN_FRONTEND=noninteractive apt-get upgrade -y --no-install-recommends \
+	&& DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+		automake \
+		cmake \
+		g++ \
+		git \
+		libtool \
+		make \
+		ninja-build \
+		python3 \
+		python \
+		wget \
+		zip \
+	&& apt-get clean \
+	&& rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+#
+# Install Bazel
+#
+RUN export BAZEL_VERSION=`cat .bazelversion` \
+        && if [ "$ARCH" = "amd64" ] ; then \
+	     curl -sfL https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-installer-linux-x86_64.sh -o bazel-${BAZEL_VERSION}-installer-linux-x86_64.sh \
+	     && chmod +x bazel-${BAZEL_VERSION}-installer-linux-x86_64.sh \
+	     && ./bazel-${BAZEL_VERSION}-installer-linux-x86_64.sh \
+	     && mv /usr/local/bin/bazel /usr/bin \
+	     && rm bazel-${BAZEL_VERSION}-installer-linux-x86_64.sh ; \
+           elif [ "$ARCH" = "arm64" ] ; then \
+             apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y curl unzip zip build-essential openjdk-11-jdk && \
+             echo "deb https://download.opensuse.org/repositories/home:/mrostecki:/bazel/xUbuntu_20.04/ /" > /etc/apt/sources.list.d/bazel.list && \
+             curl -L https://download.opensuse.org/repositories/home:/mrostecki:/bazel/xUbuntu_20.04/Release.key | apt-key add - && \
+             apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y bazel ; \
+           fi
+ 
+
+#
+# Build and keep the cache
+#
+RUN make BAZEL_BUILD_OPTS=--jobs=5 PKG_BUILD=1 ./bazel-bin/cilium-envoy && rm ./bazel-bin/cilium-envoy
+
+#
+# Absolutely nothing after making envoy deps!
+#
--- a/Dockerfile.multi_arch
+++ b/Dockerfile.multi_arch
@@ -1,0 +1,30 @@
+#
+# Cilium incremental build. Should be fast given builder-deps is up-to-date!
+#
+# cilium-builder tag is the Git SHA of the compatible build image commit.
+# Keeping the old images available will allow older versions to be built
+# while allowing the new versions to make changes that are not backwards compatible.
+#
+# The base image here should be multi-arched beforehand when used on non-amd64 platforms, such as arm64
+# platform
+#
+FROM quay.io/cilium/cilium-envoy-builder-dev:bb7c198879e48ac01ab77e0f06084a53a386d9ca as builder
+LABEL maintainer="maintainer@cilium.io"
+
+WORKDIR /go/src/github.com/cilium/cilium/envoy
+COPY . ./
+
+ARG V
+#
+# Please do not add any dependency updates before the 'make install' here,
+# as that will mess with caching for incremental builds!
+#
+RUN ./tools/get_workspace_status
+RUN make BAZEL_BUILD_OPTS=--jobs=2 PKG_BUILD=1 V=$V DESTDIR=/tmp/install cilium-envoy install
+
+#
+# Extract installed cilium-envoy binaries to an otherwise empty image
+#
+FROM scratch
+LABEL maintainer="maintainer@cilium.io"
+COPY --from=builder /tmp/install /
--- a/Makefile
+++ b/Makefile
@@ -33,13 +33,17 @@ CLANG ?= clang
 CLANG_FORMAT ?= clang-format
 BUILDIFIER ?= buildifier
 STRIP ?= $(QUIET) strip
-
-# istio_proxy builds require a version number to be set below
-ISTIO_VERSION =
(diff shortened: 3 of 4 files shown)


joestringer:
From Cilium meeting:

Our automated systems for generating docker images are not well flesh out right now. For the moment we compile an image on a developer host and push them into quay.io for consumption from main Cilium Dockerfile.

Dockerfile.builder.arm64
```diff
@@ -0,0 +1,91 @@
...
+
+#
+# Install Bazel
+#
+
+RUN apt-get update && apt-get install -y unzip zip curl
+
+
+RUN export BAZEL_VERSION=`cat BAZEL_VERSION` && echo $BAZEL_VERSION && \
+     curl -sfL https://github.com/Tick-Tocker/bazel-arm64/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-linux-aarch64 \
```
joestringer, line 16: I have the same concern here as for the other PR:

https://github.com/cilium/proxy/pull/18#discussion_r408480935

Whatever we do to resolve this, we should make it consistent across the repository.

TrevorTaoARM: @joestringer Thank you. 
I had replied this in https://github.com/cilium/proxy/pull/18#discussion_r408480935.
In summary, it's a concise and workable solution here. In case it's not workable in the future, we can still deal with it by somthing like "build from source" or others. 
Actually I had tested a "build from source" solution for this Dockerfile and it's also ok to use, but just want to get a concise and quick way here for Bazel arm64.

aanm: @TrevorTaoARM me and Joe will meet this week to discuss this. We will provide feedback.

Dockerfile.arm64
```diff
@@ -0,0 +1,26 @@
+#
+# Cilium incremental build. Should be fast given builder-deps is up-to-date!
+#
+# cilium-builder tag is the Git SHA of the compatible build image commit.
+# Keeping the old images available will allow older versions to be built
+# while allowing the new versions to make changes that are not backwards compatible.
+#
+#FROM quay.io/cilium/cilium-envoy-builder:c31482c3e49670980c05cafc914320f7949b266f-arm64 as builder
+FROM quay.io/cilium/cilium-envoy-builder:latest-arm64 as builder
```
aanm, line 9: @TrevorTaoARM Can you create a `Dockerfile.multi-arch` that could accept an ARCH as an argument? Later on we would deprecate `Dockerfile` and replace it with the `Dockerfile.multi-arch`

Dockerfile.builder.arm64
aanm: @TrevorTaoARM Can you use a bazel package for arm64, if available, from ubuntu? Or if we only need the binary would the package from open suse work? cc @mrostecki 

Also, can you create a `Dockerfile.builder.multi-arch` that could accept an ARCH as an argument? Later on we would deprecate `Dockerfile.builder` and replace it with the `Dockerfile.builder.multi-arch`

mrostecki: I can create a deb package for Debian and Ubuntu on openSUSE Build Service, based on the RPM spec we have. Give me one day, I will ping you once it's ready and post instructions how to install it.

I'm also about to submit an update to Bazel 3.0 today. :)

joestringer: Thanks for the offer @mrostecki , I think that this should resolve the remaining concerns:
* Fast turn-around to unblock other work
* Low long-term maintenance by reusing work openSUSE folks are already doing
* Easier to validate the integrity of the build chain

TrevorTaoARM: @aanm @mrostecki @joestringer Thank you.
Yes, if the Bazel binary/pkg is available (soon) from ubuntu or openSUSE, I would use it. 
I would also try to create a Dockerfile.builder.multi-arch with an ARCH argument. It would take all platforms together so it may need some changes on current work.

Dockerfile.arm64
TrevorTaoARM: @aanm Yes, I would do it also.

TrevorTaoARM:
@aanm @joestringer 
As required, I updated the PR with a multi-arch enabled Dockerfile.multi-arch and Dockerfile.builder.multi-arch, the correponding parts in the Makefile are also updated and fat-manifest support for the images is also added.
Currently, I have not found the Bazel arm64 binaries/deb pkgs from the link https://build.opensuse.org/package/binaries/devel:tools:building/bazel2.0/openSUSE_Factory_ARM
which are mentioned in https://github.com/cilium/proxy/pull/18 and thought to be feasible for us to use. If it's actually available there, I would update the Bazel installation part in the Dockerfile.builder.multi_arch correspondingly.

Dockerfile.builder.multi_arch
```diff
@@ -6,8 +6,14 @@
 #
 # Using cilium-builder as the base to ensure libc etc. are in sync.
 #
-FROM quay.io/cilium/cilium-builder:2020-02-26 as builder
+# The base image here should be multi-arched beforehand when used for non-amd64
+# platforms, e.g, arm64 platform.
```
aanm, line 10: Please leave the `Dockerfile.builder` untouched for now as we will transition from one to the other slowly. In the end it is intended that this `Dockerfile.builder.multi_arch` will replace the `Dockerfile.builder` but for now it is important that we don't brake existing builds on docker hub.

Dockerfile.multi_arch
```diff
@@ -5,10 +5,15 @@
 # Keeping the old images available will allow older versions to be built
 # while allowing the new versions to make changes that are not backwards compatible.
 #
+# The base image here should be multi-arched beforehand when used on non-amd64 platforms, such as arm64
+# platform
```
aanm, line 9: The same comment as left for the `Dockerfile.builder`, please keep the `Dockerfile` untouched and this file will be a new

mrostecki:
Sorry, getting it work for Ubuntu/Debian is harder than I though. I have my playground here:

https://build.opensuse.org/package/show/home:mrostecki:bazel/bazel

Dockerfile.multi_arch
TrevorTaoARM: @aanm OK. I would keep the current Dockerfile and Dockerfile.builder unchanged.

Dockerfile.multi_arch
```diff
@@ -0,0 +1,30 @@
...
+# Cilium incremental build. Should be fast given builder-deps is up-to-date!
+#
+# cilium-builder tag is the Git SHA of the compatible build image commit.
+# Keeping the old images available will allow older versions to be built
+# while allowing the new versions to make changes that are not backwards compatible.
+#
+# The base image here should be multi-arched beforehand when used on non-amd64 platforms, such as arm64
+# platform
+#
+FROM quay.io/cilium/cilium-envoy-builder:bb7c198879e48ac01ab77e0f06084a53a386d9ca as builder
```
aanm, line 11: I've created the `-dev` repository so this base image can be changed for
`quay.io/cilium/cilium-envoy-builder-dev:bb7c198879e48ac01ab77e0f06084a53a386d9ca`. Once we have the bazel for arch64 and the github actions we will be able to create a docker image for `quay.io/cilium/cilium-envoy-builder-dev` that has support for arch64 and amd64. :tada:

TrevorTaoARM:
@aanm I updated the Dockerfile.multi_arch and Makefile as you indicated, thanks.

mrostecki:
@aanm said that we are going to update the base image to Ubuntu 20.04.

Now we have Bazel aarch64 package on OBS:

https://build.opensuse.org/package/show/home:mrostecki:bazel

Once it's done, Bazel can be installed with the following commands:

```
echo "deb https://download.opensuse.org/repositories/home:/mrostecki:/bazel/xUbuntu_20.04/ /" > /etc/apt/sources.list.d/bazel.list"
curl -L https://download.opensuse.org/repositories/home:/mrostecki:/bazel/xUbuntu_20.04/Release.key | sudo apt-key add -
apt update
apt install bazel
```

Plans for future regarding Bazel packaging - I will:

1. Name the OBS project in different way (without home:mrostecki prefix).
2. Create github repos with all scripts, .spec and debian.* files and sync them to OBS, so people wanting to contribute can do that just from github.

Dockerfile.builder.multi_arch
```diff
@@ -0,0 +1,66 @@
...
+#
+RUN export BAZEL_VERSION=`cat .bazelversion` \
+        && if [ "$ARCH" = "amd64" ] ; then \
+	     curl -sfL https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-installer-linux-x86_64.sh -o bazel-${BAZEL_VERSION}-installer-linux-x86_64.sh \
+	     && chmod +x bazel-${BAZEL_VERSION}-installer-linux-x86_64.sh \
+	     && ./bazel-${BAZEL_VERSION}-installer-linux-x86_64.sh \
+	     && mv /usr/local/bin/bazel /usr/bin \
+	     && rm bazel-${BAZEL_VERSION}-installer-linux-x86_64.sh ; \
+           elif [ "$ARCH" = "arm64" ] ; then \
+             apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y curl unzip zip build-essential openjdk-11-jdk && \
```
mrostecki, line 52: `bazel` packages on OBS are available for x86_64/amd64 too. Maybe let's install them for every architecture and not use the official script? What do you think?

TrevorTaoARM:
@mrostecki Thank you. I think it's ok to install the Bazel package from OBS too. You could figure out the installation steps as that of arm64 and I would update it as what you give.
On the other side, I think It's also ok to keep the current installation for amd64 since maybe in the future the Bazel arm64 may also be available as that of x86_64.

joestringer:
@aanm did you intend to take another look at this PR before merging?

errordeveloper:
@TrevorTaoARM thanks for putting this together and thanks everyone for reviewing it, I think we could merge this if it would unblock manual build for someone who has Arm hardware and is happy to build natively. However, for automated builds the images will need to be cross-compiled, and that's something I started looking into, I'll report back with my finding in the next few days.

PR #25, opened 2020-08-24
build: Add support for aarch64 cross-compilation

mrostecki:
This change adds the possibility to cross-compile Envoy for aarch64
target on host by using our Bazel toolchain.

Command used for cross-compilation:

```
bazel build \
    --incompatible_enable_cc_toolchain_resolution \
    --platforms=//bazel/platforms:aarch64_cross \
    --record_rule_instantiation_callstack \
    --define cross=aarch64 \
    //:cilium-envoy
```

or:

```
bazel build \
    --incompatible_use_cc_configure_from_rules_cc \
    --crosstool_top=//bazel/toolchains:toolchain \
    --host_crosstool_top=@bazel_tools//tools/cpp:toolchain \
    --cpu=aarch64-cross \
    --define cross=aarch64 \
    //:cilium-envoy
```

Signed-off-by: Michal Rostecki <mrostecki@opensuse.org>


errordeveloper (approved):
This looks good to me! Great job @mrostecki 🏅 👍

jrajahalme:
2nd build command fails with:

`src/main/tools/linux-sandbox-pid1.cc:427: "execvp(/usr/bin/aarch64-linux-gnu-gcc, 0x430cd0)": No such file or directory`

while the 1st one seems to be building fine. What is the difference between these two build commands? Is the 1st one downloading the tools and the 2nd one uses tools from the host (which I do not have)? Anything else?

mrostecki:
@jrajahalme In order to be able to cross-compile, you need to have tools for cross-compilation with ´aarch64-linux-gnu` prefix like: `aarch64-linux-gnu-gcc`, `aarch64-linux-gnu-ld`, `aarch64-linux-gnu-ar` etc. They are required by the cross-aarch64 toolchain.

You can either:

1. Install them on your host:

```
apt install binutils-aarch64-linux-gnu crossbuild-essential-arm64 g++-aarch64-linux-gnu gcc-aarch64-linux-gnu pkg-config-aarch64-linux-gnu
```

2. Cross-compile in Docker:

```
docker run \
    --volume "$(pwd):/src" --volume "$(pwd)/_cache:/root/.cache" --workdir "/src" --rm -ti \
   cilium/image-compilers:32a185376865d6cb0f82d7339e311d998ab859c1 \
   bazel build \
    --incompatible_use_cc_configure_from_rules_cc \
    --crosstool_top=//bazel/toolchains:toolchain \
    --host_crosstool_top=@bazel_tools//tools/cpp:toolchain \
    --cpu=aarch64-cross \
    //:cilium-envoy
```

Although it's weird that the 1st command worked for you. Both commands should require tools on the host. There should be no practical difference between those two commands, apart from the fact that the 2nd one is the older method, which is still used in the most of C++ projects using Bazel, but it might get deprecated at some point. The 1st command is quite fresh and not that commonly used thing. Tensorflow, last time I checked, was using still the 2nd command.

More on difference between `--platforms` and `--crosstool_top` here: https://docs.bazel.build/versions/master/platforms-intro.html

jrajahalme:
The build on cilium dev VM running on OSX using:
```
bazel build     --incompatible_use_cc_configure_from_rules_cc \
    --extra_toolchains=//bazel/toolchains:gcc_aarch64_cross_toolchain \
    --platforms=//bazel/platforms:aarch64-cross \
    //:cilium-envoy
```

produced an amd64 binary, not ARM:
```
$ readelf -h -A -W ./bazel-bin/cilium-envoy
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00 
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              DYN (Shared object file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0x42cfe0
  Start of program headers:          64 (bytes into file)
  Start of section headers:          186574880 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         10
  Size of section headers:           64 (bytes)
  Number of section headers:         40
  Section header string table index: 39
```

mrostecki:
@jrajahalme Thanks for finding out the issue and sorry for that!

Now it should work, but I also changed both commands a bit. The changes are:

- Adding `--record_rule_instantiation_callstack` to the first command - without it, our custom cross-aarch64 toolchain gets ignored.
- Adding `--define cross=aarch64` to both commands. With the first command, there is no straighforward way to define a `config_setting` - config_settings cannot be defined based on the platform. So I decided to add a `config_setting` based on that `define` value. It's all done in the Envoy patch - https://github.com/mrostecki/proxy/blob/cross-aarch64/patches/cross-aarch64.patch

mrostecki:
Now, after both commands, I get:

```
root@a94352a241ec:/src# readelf -h -A -W ./bazel-bin/cilium-envoy
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 03 00 00 00 00 00 00 00 00 
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - GNU
  ABI Version:                       0
  Type:                              DYN (Shared object file)
  Machine:                           AArch64
  Version:                           0x1
  Entry point address:               0x4a66fc
  Start of program headers:          64 (bytes into file)
  Start of section headers:          195692128 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         10
  Size of section headers:           64 (bytes)
  Number of section headers:         34
  Section header string table index: 33
```

(in the `image-compilers` container)

jrajahalme (approved):
LGTM

PR #33, opened 2020-12-04
Istio 1.6.14

jrajahalme:
Update Istio proxy integration to release 1.6.14.

Sidecar redirection had to be changed to REDIRECT instead of TPROXY, as TPROXY redirection in upstream Istio has been broken. Fixing it would take considerable investment into integration testing (so that it would not be broken again). Istio currently always configures the original destination listener filter, which allows Cilium filter to see the original destination, which was the functional requirement to use TPROXY in the first place.

Istio seems to be confused about which listeners are ingress vs. egress filters. Fix this by comparing to the local IP address. While this works for now, this may need revisiting when Istio supports dual-stack (IPv4 and IPv6) PODs.


jrajahalme:
Corresponding CIlium PR: https://github.com/cilium/cilium/pull/14271

cilium/bpf_metadata.cc
```diff
@@ -167,22 +167,35 @@ bool Config::getMetadata(Network::ConnectionSocket& socket) {
     dip = socket.localAddress()->ip();
   }
 
+  // Sidecars may get incorrect "ingress" config, override.
```
aditighag, line 170: We don't seem to be validating the read value at L174 before it's overridden?

aditighag: Also, I'm not able to comment on the lines that are not part of the change.
But are L154-163 still valid?

jrajahalme:
Closing as obsolete

PR #36, opened 2021-01-06
proxylib: Reset the need_bytes_ after requested input_len being fulfilled

rueian:
Hi, 
While I am writing my own golang share module under proxylib, I find it might be a bug that the need_bytes_ isn't reset across message boundaries.

The problem I encountered is my golang share module will not be called if the current input length is smaller than the need_bytes_  which is requested in the previous boundary.

So I think the need_bytes_ should be reset once the input length is fulfilled and not carry to next boundary.

### DIFF
--- a/cilium/proxylib.cc
+++ b/cilium/proxylib.cc
@@ -196,6 +196,7 @@ FilterResult GoFilter::Instance::OnIO(bool reply, Buffer::Instance& data, bool e
   if (input_len < dir.need_bytes_) {
     return FILTER_OK;
   }
+  dir.need_bytes_ = 0;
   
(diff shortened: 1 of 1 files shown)


rueian:
Hi @jrajahalme, could you review this for me?

jrajahalme:
Thanks for submitting this fix. `need_bytes_` was indeed not reset anywhere, and this is the spot where it should be done, right after verifying that we received the needed number of bytes. I'll look into adding a test case to make sure there is no future regression on this.

One note though, having received the needed bytes does not necessarily signify a frame boundary on input, as a naïve parser might always ask one more byte.

jrajahalme (approved):

PR #39, opened 2021-02-09
Multi-arch cross-compilation support in GitHub actions

jrajahalme:
Builds on top of #38 (please skip 7 first commits, will rebase when #38 has merged).

- Clean up the docker build context by ignoring all files not needed for docker build (incl. .git)
- Use buildx for multi-arch and/or cross-compile builds
- Add support for docker caching
- Add support for Bazel remote caching
- Separate the builder base image (build tools) and Bazel build artifacts into separate images to allow both native and cross builds for arm64
- Fix up arm64 cross compile toolchain for older GCC versions
- Update build documentation (`README.md`)
- Add make targets for testing in docker build
- Add github actions for CI builds & testing, as well as CD on master merge


.github/workflows/build.yaml
```diff
@@ -0,0 +1,56 @@
...
+          context: .
+          file: ./Dockerfile
+          platforms: linux/amd64,linux/arm64
+          build-args: |
+            BUILDER_IMAGE=quay.io/cilium/cilium-envoy-builder:release-archive-latest
+            BAZEL_BUILD_OPTS=--remote_upload_local_results=false
+          cache-from: type=local,src=/tmp/buildx-cache
+          cache-to: type=local,dest=/tmp/buildx-cache,mode=max
+          push: true
+          tags: quay.io/${{ github.repository_owner }}/cilium-envoy-dev:${{ github.event.pull_request.head.sha }}
```
aanm, line 56: I would just add a final step to print the SHA of the images created. Something like
```yaml

      - name: Image Release Digest
        shell: bash
        run:
          echo "Digests:"
          echo "quay.io/${{ github.repository_owner }}/cilium-envoy-dev:${{ github.event.pull_request.head.sha }}@${{ steps.docker_build_ci.outputs.digest }}"
```

.github/workflows/cache-build.yaml
```diff
@@ -0,0 +1,141 @@
...
+          file: ./Dockerfile
+          platforms: linux/amd64,linux/arm64
+          build-args: |
+            BAZEL_BUILD_OPTS=--remote_upload_local_results=false
+            BUILDER_IMAGE=quay.io/${{ github.repository_owner }}/cilium-envoy-builder:release-archive-latest
+          cache-to: type=local,dest=/tmp/buildx-cache,mode=max
+          push: true
+          tags: |
+            quay.io/${{ github.repository_owner }}/cilium-envoy:latest
+            quay.io/${{ github.repository_owner }}/cilium-envoy:${{ github.sha }}
```
aanm, line 141: same for the sha. (don't forget to add an ID to this step)

.github/workflows/build.yaml
jrajahalme: Added, testing again to make sure I did not mess this up :-)

.github/workflows/cache-build.yaml
jrajahalme: Same.

PR #40, opened 2021-02-27
Envoy: Update to main branch

jrajahalme:
Rebase Envoy to the current upstream main branch. Note that this is not
a released Envoy version, so this should not be used in a Cilium release.

Most of our patches have merged upstream so get rid of them.


jrajahalme:
Superseded by #49

PR #55, opened 2021-12-01
bpf: support regular hash map for ipcache

jrajahalme:
Older kernels do not support LPM maps. Add support for hash maps so
that ipcache lookups also work for the older kernels and the fallback
syncing ipcache over xDS can be avoided.


jrajahalme:
Turns out this is not as simple, as ipcache lookups from non-LPM maps would need explicit configuration of the used prefix lengths, or upto 128 lookups. We do not want to do either.

PR #57, opened 2022-01-10
Envoy 1.21.0

jrajahalme:
- Update Envoy to release 1.21.0
- Use stable Cilium image for proxylib tests. This allows integration tests to run also on arm64.
- Add a toolchain for cross-compiled amd64 builds (amd64 build on arm64 host).
- Update build to use the same toolchain for native and cross builds, (e.g., native amd64 vs. amd64 on arm64). This allows cross-compiled build archives to be used on native builds, and natively build archives on cross-compiled builds.
- Remove Bazel platform definitions and revert to crosstool_top & cpu args. This was necessary to enable the defined toolchain to be used for native builds as well as cross-compiled builds.
- Use faster linker (gold vs ld)

Tests still only run on the local/native CPU due to Bazel issue that has been fixed but not released yet (https://github.com/bazelbuild/bazel/pull/14391). When this is fixed the goal is to run the tests for both platforms in the CI


jrajahalme:
1.21 update commits merged separately, upstream callback in another PR.

Issue #61, opened 2022-03-23
Socket re-use in proxy causing EADDRNOTAVAIL


jmcshane:
We are using https://docs.cilium.io/en/stable/gettingstarted/tls-visibility/ in our Kubernetes clusters to restrict outbound network access via L7 filters. We have a Kubernetes pod that uses [pyarrow S3FileSystem](https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html) to access data from S3 into a python computation. This library has a large upfront cost, sending a large number of requests to S3 endpoints when it is initialized.

I believe that we have found a bug in this library due to this socket re-use. Around 1% of our requests are failing regularly, and we can also get into a state where no requests successfully get through S3. When disabling the cilium L7 filters, these errors never occur.

In order to debug this issue, we started cilium-agent with `--debug-verbose=envoy --envoy-log=/tmp/envoy.log` to see the proxy debug logs. Within this log, it is clear that the 503s are returned from envoy immediately after it receives an EADDRNOTAVAIL on a specific socket.

```
[2022-03-23 12:53:23.323][401][debug][pool] creating a new connection
[2022-03-23 12:53:23.323][401][trace][filter] Set socket (137) option SO_MARK to 32030b00 (magic mark: b00, id: 12803, cluster: 0), src: 10.0.43.18:37372
[2022-03-23 12:53:23.323][401][debug][client] [C29] connecting
[2022-03-23 12:53:23.323][401][debug][connection] [C29] connecting to 52.217.196.218:443
[2022-03-23 12:53:23.323][401][debug][connection] [C29] immediate connect error: 99
[2022-03-23 12:53:23.323][401][trace][pool] not creating a new connection, shouldCreateNewConnection returned false.
[2022-03-23 12:53:23.323][401][trace][connection] [C21] socket event: 2
[2022-03-23 12:53:23.323][401][trace][connection] [C21] write ready
[2022-03-23 12:53:23.323][401][trace][connection] [C29] socket event: 3
[2022-03-23 12:53:23.323][401][debug][connection] [C29] raising immediate error
[2022-03-23 12:53:23.323][401][debug][connection] [C29] closing socket: 0
[2022-03-23 12:53:23.323][401][trace][connection] [C29] raising connection event 0
[2022-03-23 12:53:23.323][401][debug][client] [C29] disconnect. resetting 0 pending requests
[...]

PR #63, opened 2022-03-29
cilium: Do not use original source address for locally allocated destination identities

jrajahalme:
Identities for CIDR or FQDN policy based destinations are allocated
locally. These destinations are outside of the cluster and will see the
source address as masqueraded by the local node. This being the case do
not bother about using the original source address for these
destinations.

### DIFF
--- a/cilium/bpf_metadata.cc
+++ b/cilium/bpf_metadata.cc
@@ -229,13 +229,14 @@ bool Config::getMetadata(Network::ConnectionSocket& socket) {
   // Use original source address with L7 LB for local endpoint sources as policy
   // enforcement after the proxy depends on it. Otherwise
   // only use the original source address if permitted and the destination is not
-  // in the same node and is not classified as WORLD.
+  // in the same node, is not a locally allocated identity, and is not classified as WORLD.
   //
   // NOTE: Both of these options (egress_mark_source_endpoint_id_ and
   // may_use_original_source_address_) are only used for egress, so the local
   // endpoint is the source, and the other node is the destination.
   if ((egress_mark_source_endpoint_id_ && policy->getEndpointID() != 0) ||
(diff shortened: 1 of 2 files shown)


cilium/bpf_metadata.cc
```diff
@@ -229,13 +229,14 @@ bool Config::getMetadata(Network::ConnectionSocket& socket) {
...
   // only use the original source address if permitted and the destination is not
-  // in the same node and is not classified as WORLD.
+  // in the same node, is not a locally allocated identity, and is not classified as WORLD.
   //
   // NOTE: Both of these options (egress_mark_source_endpoint_id_ and
   // may_use_original_source_address_) are only used for egress, so the local
   // endpoint is the source, and the other node is the destination.
   if ((egress_mark_source_endpoint_id_ && policy->getEndpointID() != 0) ||
       (may_use_original_source_address_ &&
+       !(destination_identity & Cilium::ID::LocalIdentityFlag) &&
```
joestringer, line 239: We might get away with this today, but it could limit our future options - I have been mulling over using locally-allocated identities to identify remote nodes based on the associated labels. That said, maybe we just solve that problem another way (eg using global identity allocation for node identities).

joestringer: Not exactly sure about those semantics, but we do have a few bits to work with, so if we add this semantic to these bits then we can do something like this instead.

jmcshane:
Resolves #61 

We've tested this PR in our environment and confirmed that the changes here resolve the issue as described

joestringer (approved):

jrajahalme:
> Resolves #61
> 
> We've tested this PR in our environment and confirmed that the changes here resolve the issue as described

Thanks for the confirmation!

PR #76, opened 2022-06-16
gha: Avoid using Envoy version in builder image tag

sayboras:
### Description

This commit is to avoid using envoy version in builder image tag, mainly
to avoid chicken and egg issue if we are bumping envoy version.

Signed-off-by: Tam Mach <tam.mach@cilium.io>

### Note

This is step 1 of 2 phase commit, the next PR will move all builds to new tag, after this PR is merged.


jrajahalme (approved):
Thanks!

I added the new tags into quay manually so that the build does not need to start from scratch after this change.

PR #81, opened 2022-08-19
upstream: Set SO_REUSEPORT when binding to a specific source address

jrajahalme:
Set `SO_REUSEPORT` option to hopefully get rid of `EADDRNOTAVAIL` errors when binding to the original source address of a source pod.


jrajahalme:
TLS CI tests are failing due to expired certificates. It is not really feasible to update them other than waiting for Envoy upstream to update the certs in `release-1.21` branch. Updating to Envoy 1.22 might be an alternative.

jrajahalme:
Issued upstream Envoy update for the expired certs. We'll see the TLS tests failing until it is merged and we pick it up.

jrajahalme:
CI on `cilium/cilium` confirms that this change does not break any existing tests, so merging.

PR #82, opened 2022-08-25
Build improvements

jrajahalme:
Improvements mainly for local builds:

1. Use manual build stamping to force final binary linking even when no source files changed. This is needed to always make the version reported by Envoy match the SHA of the git repository, as required for Cilium Envoy version check. Modify make targets to also depend on the generated `SOURCE_VERSION`
2. Fix references to build archives also in Makefiles to mirror the change for version independent references on GH action workflows.
3. Remove unused docker cache references by default. Trying to fetch the non-existing cache took some valuable build time.
4. Add ~/go/bin to PATH so that API build works as expected. 
5.  Speed up builds by using the default docker syntax. This avoids the check/pull of docker/dockerfile:1.2 in the beginning of the build, saving 10-30 seconds in the beginning of the build.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -7,7 +7,7 @@
 import %workspace%/envoy.bazelrc
 
 # Overrides workspace_status_command
-build --workspace_status_command=tools/get_workspace_status
+build --workspace_status_command="bash tools/get_workspace_status"
 build:remote --remote_timeout=7200
 
 # Enable path normalization by default.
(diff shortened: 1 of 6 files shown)


sayboras (commented):
The below two failures seem to related to our cilium patches. This could be from recent commits, let me run bisect locally now.

```
#16 13589.2 //tests:cilium_tls_http_integration_test                                 FAILED in 1.2s
#16 13589.2   /root/.cache/bazel/_bazel_root/5f5d3cf254342807836215c78e76aef5/execroot/cilium/bazel-out/k8-fastbuild/testlogs/tests/cilium_tls_http_integration_test/test.log
#16 13589.2 //tests:cilium_tls_tcp_integration_test                                  FAILED in 11.0s
```

sayboras (approved):

PR #84, opened 2022-08-25
l7policy: Log requests also when not allowed, or connection fails.

jrajahalme:
Since we removed request headers from response logs we have missed logging the request headers for failed requests, while it would be useful to see exactly what was denied or what kind of a request failed.

Change to always log requests. A request will be logged as denied if a local 403 response for it was generated. This happens  for policy denies and in some misconfiguration cases (where the upstream callback returns `false`). For any other local failures (such as 503s, etc.) the request will be logged but not as "denied". The following response log will contain the status code in all cases.

This is more logical as it is the request that gets denied, not the response. [Cilium already maps denied logs to request type](https://github.com/cilium/cilium/blob/v1.12/pkg/envoy/accesslog.go#L56-L72), but before this change those logs contained response headers

### DIFF
--- a/cilium/l7policy.cc
+++ b/cilium/l7policy.cc
@@ -147,7 +147,8 @@ Http::FilterHeadersStatus AccessFilter::decodeHeaders(
     uint32_t destination_port = dip->port();
     uint32_t destination_identity = option->resolvePolicyId(dip);
 
-    // Fill in the log entry
+    // Update the log entry with the chosen destination address and current headers, as remaining
+    // filters and/or upstream may have altered headers.
     log_entry_.UpdateFromRequest(destination_identity, dst_address, headers);
 
     const auto& policy = option->getPolicy();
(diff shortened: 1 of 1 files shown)


cilium/l7policy.cc
```diff
@@ -176,13 +177,28 @@ Http::FilterHeadersStatus AccessFilter::decodeHeaders(
...
+      logType = ::cilium::EntryType::Denied;
+      config_->stats_.access_denied_.inc();
+    }
+    config_->Log(log_entry_, logType);
   }
-  config_->Log(log_entry_, logType);
+
+  // Log the response
+  log_entry_.UpdateFromResponse(headers, config_->time_source_);
+  config_->Log(log_entry_, ::cilium::EntryType::Response);
```
chancez, line 200: We now have two logs in the case of a not-allowed  request here. Should this be the case?

chancez: Ah NVM. Actually I just realized that was the point. I didn't notice they were logging different values. We always logged response, now we also appropriately log the request.

cilium/l7policy.cc
```diff
@@ -176,13 +177,28 @@ Http::FilterHeadersStatus AccessFilter::decodeHeaders(
...
+  // responses.
+  if (!allowed_) {
+    // Request was not yet logged. Log it so that the headers get logged.
+    // Default logging local errors as "forwarded".
+    // The response log will contain the locally generated HTTP error code.
+    auto logType = ::cilium::EntryType::Request;
+
+    if (headers.Status()->value() == "403") {
+      // Log as a denied request.
+      logType = ::cilium::EntryType::Denied;
```
chancez, line 192: Should this actually be `EntryType::Denied`? Will that result in an L7 response not coming through? I only see `EntryType::Denied` in one other place: https://github.com/cilium/proxy/blob/bd81a8434f5ffac11bcda6ddb0bbd7d8d68d273c/cilium/network_filter.cc#L241 and I wonder if we should be translating L7 concepts like 403 to "Denied" at all here.

chancez (commented):
The general idea makes sense, but I have an question about if we should be introspecting the status at all, or just pass it through as-is.

cilium/l7policy.cc
chancez: Ah, I missed the link in the PR description to mapping the denied requests to accesslog.TypeRequest. Then this seems like it could be logical. Does this mean 403 from upstream servers will show up as denied, even though the flow actually went through?

Eg: I have an app that `curls` GitHub.com's API and gets 403. Does that mean this L7 flow will show as denied? Technically it's forwarded, and I'm just seeing 403.

jrajahalme:
> The general idea makes sense, but I have an question about if we should be introspecting the status at all, or just pass it through as-is.

Status is being passed on as-is, it is used here to derive record type (drop/forward from policy perspective). Policy deny in the code above causes the callback to return `false`, which causes a local 403 response. 403's from upstream will not be recorded as drops as in that case policy had accepted the request and `allowed_` is `true`. 403 will be returned also for any error condition in the upstream callback... I'm now wondering if we should change the callback return value to a HTTP status code so that if `0` is returned the request is forwarded, otherwise a local reply with the given status code would be returned? That would allow the callback to return local `503` on cases when there is no destination address etc.

The upstream callback is a feature we are adding in a patch, so we are free to change it for the better :-)

chancez:
> which causes a local 403 response

That's the bit that clears everything up.

As far as the other points you made: I'm not sure I have enough background to say if that's needed or a good idea. I think if it allow future proofing interfaces while still allowing us to achieve the same results, then that it could be a wise choice.

chancez (approved):
Tentative approval. If you think refactoring the callback makes sense, maybe we should do that in a separate PR.

jrajahalme:
@chancez We can sit on the upstream callback design a bit longer. Should revisit before upstreaming it, though.

PR #95, opened 2022-12-05
Enable OAuth2 and JWT envoy filters

### LINKED ISSUE #93: Support envoy.filters.http.oauth2
OAuth2 is a pretty common way to secure public facing websites/endpoints, and it would be ideal if using Cilium Ingress with a custom CEC, if we could configure the oauth2 filter to handle authentication.

### LINKED ISSUE #94: Support envoy.filters.http.jwt_authn
Similar to #93. JWTs are a pretty common way to secure internal and public facing websites/endpoints, and it would be ideal if using Cilium Ingress with a custom CEC, if we could configure the JWT filter to handle authentication.

chancez:
Closes #93
Closes #94


jrajahalme (approved):
Thanks!

PR #97, opened 2022-12-09
bpf_metadata: Filter out connections with invalid socket mark

jrajahalme:
Filter out unwanted traffic by checking the socket mark for the value set by Cilium datapath.


sayboras:
This PR seems to be stale, let's close it.

PR #99, opened 2023-01-20
policy: Add support for port ranges

jrajahalme:
Add new end_port field to NPDS proto and implement support in cilium/network_policy.cc.

Check that wildcard port entires do not have port range as overlapping port ranges are not supported. Otherwise we could remove the special 0 port lookup, if 0-65535 range was inserted.

Change the port parameter in C++ functions to uint16_t, even though protobuf has uint32, as there is no uint16 type in protobufs.

Add the new field to existing tests in various configurations.


sayboras (approved):
Review mainly with the test cases, lgtm :100: 

Some of validation can be simplified or even removed, but it depends on how everything is plumbed together, looking forward to checking related changes in cilium/cilium.

PR #106, opened 2023-02-12
Cross-compilation enhancements

jrajahalme:
Use Bazel platforms for (cross-compilation) toolchain resolution. This allows removal of a BoringSSL build patch, as the upstream assumes bazel platforms are used for toolchain resolution, and we had patched BoringSSL build to allow for legacy CPU selection.

We still need to keep `cc_toolchain_suite` in `bazel/toolchains/BUILD` and `crosstool_top` in `bazel/platform_mappings` (new) to work around a bazel issue where the synthetic `local_config_cc` is still used causing a cross-compilation failure.

Add support for cross-compiling and "cross-running" tests. GH actions are changed to build test archive for both amd64 and arm64, but running the foreign architecture tests is not yet done automatically. Running the tests for both targets can be done like so:
```
$ ARCH=multi make docker-tests
```
This requires Docker `binfmt_misc` support as provided by Docker Desktop or as installed on regular docker via:
```
docker run --privileged --rm tonistiigi/binfmt --install all
```
(See https://docs.docker.com/build/building/multi-platform/).

Change docker build to use buildx by default, and add local cache mount to the new "runner" stage to speed up local docker test runs.


patches/0007-boringssl-Compat-with-cross-compile-using-legacy-cpu.patch
```diff
@@ -62,65 +62,15 @@ index 0000000000..f5b95af7e7
 +-- 
 +2.36.0
 +
-diff --git a/bazel/0002-bazel-Use-old-stype-cpu-flag-to-detect-target-platfo.patch b/bazel/0002-bazel-Use-old-stype-cpu-flag-to-detect-target-platfo.patch
```
sayboras, line 65: it's unrelated to this PR, but I have been wondering on how these patche files can be applied/generated.

Personally, I use `git appy and git format-patch` command, but it requires some manual update.

sayboras (approved):
I have tried this PR locally with my repo, all seems working as expected :heavy_check_mark:

patches/0007-boringssl-Compat-with-cross-compile-using-legacy-cpu.patch
jrajahalme: I'll try to write a guide with the next update, but basically you'd checkout Envoy at the same SHA as referenced in the WORKSPACE file, create a new branch and then `git apply` the patches on that branch. Then you would rebase that branch to the new Envoy version, and correct any conflicts, and finally:
```
git format-patch -o ~/../../cilium/proxy/patches upstream/release/v1.23
```
(assuming the next rebase would be onto the top of the v1.23 release branch)

In practice I have the patches branch on my Envoy fork that I keep rebasing, but the above should get you there.

Note that updating patches to Envoy dependencies require taking the above process one level deeper: Check out the Envoy dependency on the same version/SHA Envoy is using, and apply the patches from envoy/bazel into them, rebase and generate new patches, but this time to the envoy/bazel directory (or wherever they were), then applying those changes to the Envoy fix-up branch and generating cilium/proxy patches from there like described above. Essentially the "correct any conflicts" step may involve this kind of work.

PR #109, opened 2023-02-13
GH actions: build from scratch and persist test logs for successful test runs

jrajahalme:
Build test and build archives from scratch on master push to avoid accumulating unused files in the build caches.
This is now possible due to the large runners and more efficient compiling with clang.

Also push test logs on successful test runs for reference. Failing runs will not complete the docker build step, so no test logs will be pushed, but those can be observed from the build logs on Github.


jrajahalme:
Will have to use `-ci` instead of `-dev` for the test logs, a moment...

jrajahalme:
Or remove the push from CI tests, as adding any privileges can be dangerous.

sayboras:
> Or remove the push from CI tests, as adding any privileges can be dangerous.

I think we can just remove it, GHA logs are available anyway.

PR #127, opened 2023-02-18
Fix local docker tests

jrajahalme:
Use correct archive image also for local docker tests launched via `make`.

Pass `NO_CACHE` to docker tests so that build cache can be cleared if it becomes corrupted.

Reintroduce jobs limit for local testing based on system memory. It appears that also with clang compiling some of the test sources takes a lot of memory, and the build was killed as too many parallel jobs exhausted system memory.

Finally, check that bazel and clang-15 are installed, and pull them automatically if needed when running `make`. This streamlines local building and testing on a fresh VM.

### DIFF
--- a/Dockerfile.tests
+++ b/Dockerfile.tests
@@ -66,8 +66,10 @@ ARG V
 ARG BAZEL_BUILD_OPTS
 ARG BUILDARCH
 ARG TARGETARCH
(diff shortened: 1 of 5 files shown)


sayboras (approved):
This is great, I copied the llvm step from docker file to install required clang version manually :medal_military:

PR #128, opened 2023-02-18
Fix format checking and apply fixes

jrajahalme:
Make targets `check` and `fix` were no longer functional. Fix them to reflect upstream Envoy changes. Unfortunately there are two incompatibilities left:

- BUILD check insists on loading envoy_build_system.bzl in the upstream format, which is the wrong format for a project using Envoy as a dependency. Due to this BUILD fixing is disabled
- WORKSPACE checks are also upstream Envoy specific, and this can not be disabled. Due to this `make check` will complain about the WORKSPACE file, but luckily `make fix` will not mess with it.

To make formatting work correctly, `.clang-format` is synced from upstream Envoy.

Use the above to fix the formatting. In addition to the trivial (line lengths, white spacing) this includes:
- Use `min_len` instead of `min_bytes` in protobuf validation
- Add NOLINT options to not warn about:
  - Namespacing in a C-header file
  - valid memcpy use
  - Use of realtime (usleep) in tests as this is currently required due to limitations on the test framework
- Replace std::unordered_set and std::unordered_map with abseil equivalents
- Use default visibility for BUILD targets
- Replace dependency on the Envoy "@envoy//source/exe:envoy_common_lib" with the minimal targets instead
- Add dependencies to @envoy_api targets used (even though build currently also works without these)


sayboras (approved):
Probably we can just skip check for 2 errors in WORKSPACE

```diff
diff --git a/WORKSPACE b/WORKSPACE
index 0d7e20d3..6f971488 100644
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -20,7 +20,9 @@ ENVOY_SHA = "9c4467263303640a81a7d0d2ee7da4eed2009369"
 
 ENVOY_SHA256 = "f9c29978f3f104431847133f7fa493c0d422f8be4467789cea2a320f25c3c867"
 
+# // clang-format off: unexpected @bazel_tools reference, please indirect via a definition in //bazel
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+# // clang-format on
 
 local_repository(
     name = "envoy_build_config",
@@ -42,7 +44,9 @@ http_archive(
     ],
     sha256 = ENVOY_SHA256,
     strip_prefix = ENVOY_REPO + "-" + ENVOY_SHA,
+    # // clang-format off: Envoy's format check: Only repository_locations.bzl may contains URL references
     url = "https://github.com/" + ENVOY_PROJECT + "/" + ENVOY_REPO + "/archive/" + ENVOY_SHA + ".tar.gz",
+    # // clang-format on
 )
 
 #
```

.clang-format
```diff
@@ -1 +1,17 @@
-BasedOnStyle: Google
+---
```
sayboras, line 1: should we get clang-tidy as well?

jrajahalme: That would be a nice improvement!

jrajahalme:
> Probably we can just skip check for 2 errors in WORKSPACE
> 
> ```diff
> diff --git a/WORKSPACE b/WORKSPACE
> index 0d7e20d3..6f971488 100644
> --- a/WORKSPACE
> +++ b/WORKSPACE
> @@ -20,7 +20,9 @@ ENVOY_SHA = "9c4467263303640a81a7d0d2ee7da4eed2009369"
>  
>  ENVOY_SHA256 = "f9c29978f3f104431847133f7fa493c0d422f8be4467789cea2a320f25c3c867"
>  
> +# // clang-format off: unexpected @bazel_tools reference, please indirect via a definition in //bazel
>  load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
> +# // clang-format on
>  
>  local_repository(
>      name = "envoy_build_config",
> @@ -42,7 +44,9 @@ http_archive(
>      ],
>      sha256 = ENVOY_SHA256,
>      strip_prefix = ENVOY_REPO + "-" + ENVOY_SHA,
> +    # // clang-format off: Envoy's format check: Only repository_locations.bzl may contains URL references
>      url = "https://github.com/" + ENVOY_PROJECT + "/" + ENVOY_REPO + "/archive/" + ENVOY_SHA + ".tar.gz",
> +    # // clang-format on
>  )
>  
>  #
> ```

Did not think this would work, but it does, thanks!

jrajahalme:
@sayboras Added one commit based on your suggestion, thanks!

PR #131, opened 2023-02-19
dev: Add Vagrant file with ubuntu

sayboras:
### Description

Currently, bazel configuration requires clang-15 in specific path (e.g. /usr/bin/clang-15), which might not work well for non-ubuntu local machine. This commit is to add Vagrantfile to facilitate local dev.

### Testing

Testing was done locally with the below command

```
$ VM_MEMORY=16000 VM_CPUS=8 vagrant up
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,8 @@
 # generated for docker builds via make
 /SOURCE_VERSION
 /.dockerignore
+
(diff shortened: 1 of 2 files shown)


jrajahalme (approved):
Thanks, this will make life a bit easier :-)

PR #132, opened 2023-02-19
tests: Add minimal protobuf for test version of bpf_metadata filter

jrajahalme:
Envoy now requires each filter to have a unique config type.


sayboras (approved):
Thanks a lot

Issue #139, opened 2023-03-02
CI: cilium_tls_http_integration_test failure


sayboras:
### Description

The below run is failed without any related change. This is mainly due to multiple jobs running in parallel, the workaround is to run with job limit to 1. So there could be overlapping resources between the tests.

```bash
bazel  test --platforms=//bazel:linux_x86_64 --config=release --jobs=1 --test_timeout=300 --test_output=errors //tests/...
```

https://github.com/cilium/proxy/actions/runs/4272080973/jobs/7439190389
[logs_1256.zip](https://github.com/cilium/proxy/files/10868372/logs_1256.zip)
```
#14 350.1 //tests:cilium_tls_tcp_integration_test                                  PASSED in 3.6s
#14 350.1 //tests:cilium_websocket_codec_integration_test                          PASSED in 2.8s
#14 350.1 //tests:cilium_websocket_decap_integration_test                          PASSED in 0.4s
#14 350.1 //tests:cilium_websocket_encap_integration_test                          PASSED in 2.6s
#14 350.1 //tests:cilium_tls_http_integration_test                                 FAILED in 10.7s
```

jrajahalme:
I've had this happen locally, file system watcher can't open socket, claims too many open files. I triaged this locally, files were in different directories, so that should not be the problem. We only use file system watcher for xDS in tests, so making the tests execute one at a time is a good workaround. Let's make sure we build them first with more parallel jobs in all cases, though.

PR #141, opened 2023-03-03
test: Limit number of jobs running test

sayboras:
This is to add local_test_jobs flag in BAZEL_TEST_OPTS for below points:

- Build is still running in parallel
- Test is running with limited local_test_jobs to avoid any potential issue mentioned in #139.

Fixes: #139

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ ifdef BAZEL_REMOTE_CACHE
   BAZEL_BUILD_OPTS += --remote_cache=$(BAZEL_REMOTE_CACHE)
 endif
 
-BAZEL_TEST_OPTS ?= --jobs=HOST_RAM*.0003 --test_timeout=300
+BAZEL_TEST_OPTS ?= --jobs=HOST_RAM*.0003 --test_timeout=300 --local_test_jobs=1
 BAZEL_TEST_OPTS += --test_output=errors
 
 BUILDARCH := $(subst aarch64,arm64,$(subst x86_64,amd64,$(shell uname -m)))


sayboras:
Github action is not running due to https://www.githubstatus.com/incidents/qfby3shb6r3w, will re-open to trigger after the incident is resolved.

Makefile
```diff
@@ -28,7 +28,7 @@ ifdef BAZEL_REMOTE_CACHE
   BAZEL_BUILD_OPTS += --remote_cache=$(BAZEL_REMOTE_CACHE)
 endif
 
-BAZEL_TEST_OPTS ?= --jobs=HOST_RAM*.0003 --test_timeout=300
+BAZEL_TEST_OPTS ?= --jobs=HOST_RAM*.0003 --test_timeout=300 --local_test_jobs=1
```
jrajahalme, line 31: Unfortunately this will also limit the builds to one job, which would be a huge waste on the big runners. We'll need to split the test runs to two steps everywhere:
1. `bazel build --build_tests_only` to build the test binaries, like in the `envoy-test-deps` make target
2. Run the tests with `bazel test --jobs=1`

On the makefiles we can make the test run target depend on the test-deps target, but we'll have to make sure they are paired properly (so that the deps are built with all the same build options as the test run).

If we are seeing this same issue in the CI, then we can add the `--jobs=1` to the test run step, assuming we have build all the test code in a separate build step.

jrajahalme (changes requested):

jrajahalme (approved):
I missed the fact that this PR is adding `local_test_jobs` rather than `jobs`, sorry!

This is exactly what we need to run the tests one by one!

PR #147, opened 2023-03-18
gha: Update workflow for v1.23 branch

sayboras:
### Description

This PR is to support multiple release branches (e.g. master, v1.23) for this repo.

### Testing

Testing was done with temp commit :

https://github.com/cilium/proxy/actions/runs/4455061365/jobs/7824640229?pr=147
https://github.com/cilium/proxy/actions/runs/4455061363/jobs/7824640220?pr=147


.github/workflows/build-envoy-images-release.yaml
```diff
@@ -2,7 +2,7 @@ name: Refresh test & build cache & build latest
 on:
   push:
     branches:
-      - master
+      - v1.23
```
jrajahalme, line 5: Looks like this change will stop the release builds off of the master branch?

.github/workflows/build-envoy-image-ci.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
         with:
           path: /tmp/buildx-cache
           key: docker-cache-${{ github.head_ref }}
-          restore-keys: docker-cache-master
+          restore-keys: docker-cache-v1.23
```
jrajahalme, line 20: Not sure how significant this is, but as docker caching is just a perf optimization, maybe we could just keep one cache, for whatever build was the latest? Typically that would be for a PR on top of master, and it should not hurt too much if docker cache is not effective when building for a release branch?

.github/workflows/build-envoy-image-ci.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
       - name: Prep for build
         run: |
           echo "${{ github.event.pull_request.head.sha }}" >SOURCE_VERSION
-          echo "ENVOY_VERSION=$(cat ENVOY_VERSION)" >> $GITHUB_ENV
+          echo "ENVOY_MINOR_RELEASE=$(cat ENVOY_VERSION | sed 's/envoy-\([0-9]\+\.[0-9]\+\)\..*/v\1/')" >> $GITHUB_ENV
```
jrajahalme, line 37: Looks like we did not use `ENVOY_VERSION` for anything... so OK for removing it.

.github/workflows/build-envoy-image-ci.yaml
```diff
@@ -67,6 +67,17 @@ jobs:
           echo "Digests:"
           echo "quay.io/${{ github.repository_owner }}/cilium-envoy-dev:${{ env.BAZEL_VERSION }}-${{ env.BUILDER_DOCKER_HASH }}@${{ steps.docker_build_builder_ci.outputs.digest }}"
 
+      - name: Retrieve Archive image
+        id: cilium_archive_tag
+        shell: bash
+        run: |
+          if docker buildx imagetools inspect quay.io/${{ github.repository_owner }}/cilium-envoy-builder:${{ env.ENVOY_MINOR_RELEASE }}-archive-latest &>/dev/null; then
+            echo archive_image="quay.io/${{ github.repository_owner }}/cilium-envoy-builder:test-${{ env.ENVOY_MINOR_RELEASE }}-archive-latest" >> $GITHUB_OUTPUT
```
jrajahalme, line 75: `test-` prefix to the tag here looks like a mistake?

.github/workflows/build-envoy-image-ci.yaml
jrajahalme: Maybe we should get `ENVOY_MINOR_RELEASE` before this step to use it here?

.github/workflows/build-envoy-images-release.yaml
jrajahalme: So why not add the release branches and keep the master branch?

jrajahalme: And maybe keep `ENVOY_MINOR_RELEASE` as `master` for the master branch, as it will happen that master and a release branch will have the same Envoy minor version at some point in near future?

jrajahalme (changes requested):
Some early comments :-)

.github/workflows/build-envoy-images-release.yaml
sayboras: > Looks like this change will stop the release builds off of the master branch?

ah this PR is merging to branch v1.23, which will contain the source code for proxy 1.23.x release. It's similar to what we have with cilium/cilium repo. 

The master branch is mantained separately, required changes are done #148.

.github/workflows/build-envoy-image-ci.yaml
sayboras: good point :100:

.github/workflows/build-envoy-images-release.yaml
```diff
@@ -125,7 +128,7 @@ jobs:
       - name: Prep for build
         run: |
           echo "${{ github.sha }}" >SOURCE_VERSION
-          echo "ENVOY_VERSION=$(cat ENVOY_VERSION)" >> $GITHUB_ENV
+          echo "ENVOY_MINOR_RELEASE=$(cat ENVOY_VERSION | sed 's/envoy-\([0-9]\+\.[0-9]\+\)\..*/v\1/')" >> $GITHUB_ENV
```
jrajahalme, line 131: `ENVOY_MINOR_RELEASE` is already set as `v1.23` above, no need to set it here? Or should we make the workflow fail if `ENVOY_VERSION` file does not agree with env var `ENVOY_MINOR_RELEASE`?

.github/workflows/ci-tests.yaml
```diff
@@ -29,8 +32,8 @@ jobs:
         uses: actions/cache@6998d139ddd3e68c71e9e398d8e40b71a2f39812 # v3.2.5
         with:
           path: /tmp/buildx-cache
-          key: docker-cache-tests
-          restore-keys: docker-cache-master
+          key: docker-cache-tests-v1.23
+          restore-keys: docker-cache-v1.23
```
jrajahalme, line 36: Use `-${{ env.ENVOY_MINOR_RELEASE }}` here as well instead of hard coding the minor version?

jrajahalme (changes requested):
Minor nits, but will be good to get this right!

.github/workflows/build-envoy-images-release.yaml
jrajahalme: Right, missed that nuance.

.github/workflows/build-envoy-images-release.yaml
sayboras: ENVOY_VERSION is removed as per one of above convo. However, I have added minor release in binary version check step, so it should catch if there is any mismatch.

.github/workflows/build-envoy-image-ci.yaml
```diff
@@ -92,10 +107,11 @@ jobs:
           docker rm -fv cilium-envoy
           envoy_version=$(./cilium-envoy --version)
           echo ${envoy_version}
-          [[ "${envoy_version}" == *"${{ github.event.pull_request.head.sha }}"* ]]
+          [[ "${envoy_version}" == *"${{ github.event.pull_request.head.sha }}/${{ env.ENVOY_MINOR_RELEASE }}"* ]]
```
jrajahalme, line 110: Locally I get this output for `./cilium-envoy --version`:
```
./cilium-envoy  version: b308011020900e90579ad5927857d01f5de8016f/1.24.2/Distribution/DEBUG/BoringSSL
```
Note that there is no `v` in `1.24.2`! So the change here would fail the workflow for non-matching version.

.github/workflows/build-envoy-images-release.yaml
```diff
@@ -188,12 +191,12 @@ jobs:
           build-args: |
             BUILDER_BASE=quay.io/${{ github.repository_owner }}/cilium-envoy-builder:${{ env.BAZEL_VERSION }}-${{ env.BUILDER_DOCKER_HASH }}
             BAZEL_BUILD_OPTS=--remote_upload_local_results=false
-            ARCHIVE_IMAGE=quay.io/${{ github.repository_owner }}/cilium-envoy-builder:master-archive-latest
+            ARCHIVE_IMAGE=quay.io/${{ github.repository_owner }}/cilium-envoy-builder:${{ env.ENVOY_MINOR_RELEASE }}-archive-latest
```
jrajahalme, line 194: Now that the Envoy version and brach name may need to be different, how about using the (release) branch name here instead? So use `v1.23` here, but derive it from the branch name rather than from the Envoy version? That way we could use the same script (copied or otherwise) for multiple release branches even if they happened to be on the same Envoy version in future?

.github/workflows/build-envoy-images-release.yaml
```diff
@@ -203,10 +206,11 @@ jobs:
           docker rm -fv cilium-envoy
           envoy_version=$(./cilium-envoy --version)
           echo ${envoy_version}
-          [[ "${envoy_version}" == *"${{ github.sha }}"* ]]
+          [[ "${envoy_version}" == *"${{ github.sha }}/${{ env.ENVOY_MINOR_RELEASE }}"* ]]
```
jrajahalme, line 209: Same issue with `v` in `env.ENVOY_MINOR_RELEASE`

jrajahalme (changes requested):
There is a mismatch with the release branch name (e.g., `v1.23` and Envoy minor version as embedded in `cilium-envoy --version`, where there is no `v`. Suggest make this more flexible by automatically picking up the branch name and use it where `env.ENVOY_MINOR_RELEASE` is used currently, and then define `env.ENVOY_MINOR_RELEASE` without the `v` letter and use it in the --version check. This way we have separated envoy versioning from the release branch name so we can have different release branches even if Envoy version remained the same for some reason.

.github/workflows/build-envoy-images-release.yaml
sayboras: ah good point :100:

Issue #152, opened 2023-04-01
Add CI test with cilium-cli


jrajahalme:
Add a CI test workflow where each PR, release, and master thread is tested like this:

After `cilium/proxy` build for PR, release, or master is available, do:
- pull `cilium\cilium` master
- replace the reference to the new `cilium-envoy` into `images/cilium/Dockerfile`
- build cilium and operator images locally:
  - `DOCKER_IMAGE_TAG=test make docker-cilium-image docker-operator-generic-image`
- create a kind cluster:
  - `kind create cluster`
- load the locally built cilium images with the new `cilium-envoy`:
  - `kind load docker-image quay.io/cilium/operator-generic:test quay.io/cilium/cilium:test`
- pull latest master of `cilium/cilium-cli` & build locally
  - could also do the latest released version, but when adding new tests to cilium-cli, pulling the latest source is useful
- install cilium with `cilium-cli`, e.g. (while in the `cilium/cilium` directory)
  - `cilium install --chart-directory install/kubernetes/cilium --config monitor-aggregation=none --helm-set loadBalancer.l7.backend=envoy --helm-set tls.secretsBackend=k8s --agent-image=quay.io/cilium/cilium:test --operator-image=quay.io/cilium/operator-generic:test --helm-set image.pullPolicy=IfNotPresent --helm-set operator.image.pullPolicy=IfNotPresent --config debug=true --config debug-verbose=envoy`
   - `cilium hubble enable`
   - `cilium hubble port-forward&`
   - `cilium connectivity test`

jrajahalme:
Bonus features: Allow triggering this from the PR comments so that you can choose the branch/commit SHA of either Cilium and/or cilium-cli to test against!

mhofstetter:
@jrajahalme

are there specific reasons why we should enable and port-forward to hubble in this CI step?

sayboras:
> are there specific reasons why we should enable and port-forward to hubble in this CI step?

This is mainly for validating the hubble flow as part connectivity test.

PR #153, opened 2023-04-04
ci: introduce cilium integration test

mhofstetter:
This commit introduces a GitHub Actions workflow which executes the Cilium integration tests.

The actual integration test performs the following steps:
* checking out Cilium repo (default: master)
* installing cilium via Cilium CLI (default: latest)
* setting up a kind cluster
* patching Cilium Agent Dockerfile with Proxy image of current branch
* building and pushing the Cilium images with the modified Envoy Proxy
* performing the Cilium connectivity tests

The following events are triggering the workflow:
* push to master
* opening pull request targeting master (https://github.com/cilium/proxy/actions/runs/4616447297/jobs/8161468972?pr=153)
* commenting on pull request with the option to override versions of Cilium & Cilium CLI (`/test-cilium-integration
  [cilium=<ciliumRepoRef>] [ciliumCli=<ciliumCliImageTag>]`)

Workflow runs triggered by a comment on the pull request are currently not linked to the actual PR via its checks because this might not be desired in every situation. But the action responds to the comment by adding the URL to the workflow run and mentioning its success/failure. Re-Triggering a failed cilium integration test is also possible via UI - e.g. that should not be the goal of this "issue comment" feature.. At best it looks like this :crossed_fingers: 

screenshot-2023-04-04-15-29-42

Fixes: https://github.com/cilium/proxy/issues/152

### DIFF
--- a/.github/workflows/cilium-integration-tests.yaml
+++ b/.github/workflows/cilium-integration-tests.yaml
@@ -1,0 +1,187 @@
+name: Cilium Integration Tests
+on:
+  push:
+    branches:
+      - master
+  pull_request_target:
+    types:
+      - opened
+      - reopened
+      - synchronize
+    branches:
+      - master
+  issue_comment:
+    types:
+      - created
+env:
+  KIND_VERSION: v0.18.0
+  CILIUM_REPO_REF: master
+  CILIUM_CLI_TAG: latest
+
+jobs:
+  cilium-connectivity-tests:
+    timeout-minutes: 360
+    name: Cilium Connectivity Tests
+    if: |
+      (github.event_name == 'issue_comment' && github.event.issue.pull_request && startsWith(github.event.comment.body, '/test-cilium-integration')) ||
+      github.event_name == 'push' ||
+      github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Prepare variables for pushes to master
+        if: github.event_name == 'push'
+        run: |
+          echo "PROXY_IMAGE=quay.io/cilium/cilium-envoy" >> $GITHUB_ENV
+          echo "PROXY_TAG=${{ github.sha }}" >> $GITHUB_ENV
+
+      - name: Prepare variables for PR
+        if: github.event_name == 'pull_request'
+        run: |
+          echo "PROXY_IMAGE=quay.io/cilium/cilium-envoy-dev" >> $GITHUB_ENV
+          echo "PROXY_TAG=${{ github.event.pull_request.head.sha }}" >> $GITHUB_ENV
+
+      - name: Prepare variables for issue comment on a PR
+        if: github.event_name == 'issue_comment'
+        run: |
+          echo "PROXY_IMAGE=quay.io/cilium/cilium-envoy-dev" >> $GITHUB_ENV
+          echo "PROXY_TAG=${{ github.event.issue.pull_request.head.sha }}" >> $GITHUB_ENV
+
+          commentBody="${{ github.event.comment.body }}"
+
+          ciliumRef=${CILIUM_REPO_REF}
+          if [[ "$commentBody" == *" cilium="* ]]; then
+            ciliumRef=$(echo "$commentBody" | sed -E 's|.* cilium=([^ ]*).*|\1|g')
+          fi
+          echo "CILIUM_REPO_REF=${ciliumRef}" >> $GITHUB_ENV
+
+          ciliumCliTag=${CILIUM_CLI_TAG}
+          if [[ "${{ github.event.comment.body }}" == *" ciliumCli="* ]]; then
+            ciliumCliTag=$(echo "$commentBody" | sed -E 's|.* ciliumCli=([^ ]*).*|\1|g')
+          fi
+          echo "CILIUM_CLI_TAG=${ciliumCliTag}" >> $GITHUB_ENV
+
+      - name: Reporting start to issue comment
+        uses: actions/github-script@98814c53be79b1d30f795b907e553d8679345975 # v6.4.0
+        if: github.event_name == 'issue_comment'
+        with:
+          script: |
+            await github.rest.issues.updateComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              comment_id: ${{ github.event.comment.id }},
+              body: `${{ github.event.comment.body }}
+            
+              👋 https://github.com/${{ github.repository_owner }}/${{ github.event.repository.name }}/actions/runs/${{ github.run_id }}`
+            })
+            
+            await github.rest.reactions.createForIssueComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              comment_id: ${{ github.event.comment.id }},
+              content: 'rocket'
+            })
+
+      - name: Checkout Cilium ${{ env.CILIUM_REPO_REF }}
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
+        with:
+          repository: cilium/cilium # Be aware that this is the Cilium repository and not the one of the proxy itself!
+          ref: ${{ env.CILIUM_REPO_REF }}
+
+      - name: Install Cilium CLI ${{ env.CILIUM_CLI_TAG }}
+        run: |
+          cid=$(docker create quay.io/cilium/cilium-cli-ci:${{ env.CILIUM_CLI_TAG }} ls)
+          sudo docker cp $cid:/usr/local/bin/cilium /usr/local/bin
+          docker rm $cid
+          cilium version
+
+      - name: Create kind cluster
+        uses: helm/kind-action@d8ccf8fb623ce1bb360ae2f45f323d9d5c5e9f00 # v1.5.0
+        with:
+          version: ${{ env.KIND_VERSION }}
+          config: '.github/kind-config.yaml'
+          cluster_name: 'kind'
+
+      - name: Patch Cilium Agent Dockerfile
+        shell: bash
+        run: |
+          sed -i -E 's|(FROM )(quay\.io\/cilium\/cilium-envoy:)([0-9a-z]*)(@sha256:[0-9a-z]*)( as cilium-envoy)|\1${{ env.PROXY_IMAGE }}:${{ env.PROXY_TAG }}\5|' ./images/cilium/Dockerfile
+          cat ./images/cilium/Dockerfile
+
+      - name: Wait for Cilium Proxy image to be available
+        timeout-minutes: 30
+        shell: bash
+        run: until docker manifest inspect ${{ env.PROXY_IMAGE }}:${{ env.PROXY_TAG }} &> /dev/null; do sleep 15s; done
+
+      - name: Build Cilium Agent & Operator with patched Cilium Proxy Image
+        shell: bash
+        run: DOCKER_IMAGE_TAG=test make docker-cilium-image docker-operator-generic-image
+
+      - name: Load Cilium Images into kind
+        shell: bash
+        run: |
+          kind load docker-image \
+            --name kind \
+            quay.io/cilium/operator-generic:test \
+            quay.io/cilium/cilium:test
+
+      - name: Install Cilium
+        shell: bash
+        run: |
+          cilium install \
+            --chart-directory install/kubernetes/cilium \
+            --config monitor-aggregation=none \
+            --helm-set loadBalancer.l7.backend=envoy \
+            --helm-set tls.secretsBackend=k8s \
+            --agent-image=quay.io/cilium/cilium:test \
+            --operator-image=quay.io/cilium/operator-generic:test \
+            --helm-set image.pullPolicy=Never \
+            --helm-set operator.image.pullPolicy=Never \
+            --config debug=true \
+            --config debug-verbose=envoy
+
+          cilium hubble enable \
+            --chart-directory install/kubernetes/cilium
+
+          cilium hubble port-forward&
+
+      - name: Execute Cilium Connectivity Tests
+        shell: bash
+        run: cilium connectivity test
+
+      - name: Reporting success to issue comment
+        uses: actions/github-script@98814c53be79b1d30f795b907e553d8679345975 # v6.4.0
+        if: success() && github.event_name == 'issue_comment'
+        with:
+          script: |
+            await github.rest.reactions.createForIssueComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              comment_id: ${{ github.event.comment.id }},
+              content: '+1'
+            })
+
+      - name: Gather Cilium system dump
+        if: failure()
+        shell: bash
+        run: cilium sysdump --output-filename cilium-integration-test-sysdump
+
+
+      - name: Upload Cilium system dump
+        if: failure()
+        uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
+        with:
+          name: cilium-integration-test-sysdump
+          path: cilium-integration-test-sysdump-*.zip
+          retention-days: 5
+
+      - name: Reporting failure to issue comment
+        uses: actions/github-script@98814c53be79b1d30f795b907e553d8679345975 # v6.4.0
+        if: failure() && github.event_name == 'issue_comment'
+        with:
+          script: |
+            await github.rest.reactions.createForIssueComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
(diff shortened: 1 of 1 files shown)


.github/workflows/cilium-integration-tests.yaml
```diff
@@ -0,0 +1,155 @@
+name: Cilium Integration Tests
+on:
+  push:
+    branches:
+      - master
+  pull_request: {}
```
mhofstetter, line 6: replace `pull_request` with `pull_request_trigger` before merge

.github/workflows/cilium-integration-tests.yaml
```diff
@@ -0,0 +1,155 @@
...
+env:
+  KIND_VERSION: v0.18.0
+  CILIUM_REPO_REF: master
+  CILIUM_CLI_TAG: latest
+
+jobs:
+  cilium-connectivity-tests:
+    timeout-minutes: 360
+    name: Cilium Connectivity Tests
+    runs-on: ubuntu-latest-64-cores-256gb
```
sayboras, line 19: we can use smaller runner here (e.g. ubuntu-latest)

.github/workflows/cilium-integration-tests.yaml
```diff
@@ -0,0 +1,155 @@
...
+          echo "PROXY_TAG=${{ github.sha }}" >> $GITHUB_ENV
+
+      - name: Prepare variables for PR
+        if: github.event_name == 'pull_request'
+        run: |
+          echo "PROXY_IMAGE=quay.io/cilium/cilium-envoy-dev" >> $GITHUB_ENV
+          echo "PROXY_TAG=${{ github.event.pull_request.head.sha }}" >> $GITHUB_ENV
+
+      - name: Prepare variables for issue comment on a PR
+        if: github.event_name == 'issue_comment' && github.event.issue.pull_request && startsWith(github.event.comment.body, '/cilium-integration-test')
```
sayboras, line 34: we are kind of familiar with /test. What do you think?

.github/workflows/cilium-integration-tests.yaml
```diff
@@ -0,0 +1,155 @@
...
+      - uses: actions/github-script@v6
+        if: failure() && github.event_name == 'issue_comment' && github.event.issue.pull_request && startsWith(github.event.comment.body, '/cilium-integration-test')
+        with:
+          script: |
+            await github.rest.reactions.createForIssueComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              comment_id: ${{ github.event.comment.id }},
+              content: '-1'
+            })
```
sayboras, line 155: can we add a small step to capture sysdump in case of failure here ?

.github/workflows/cilium-integration-tests.yaml
```diff
@@ -0,0 +1,155 @@
...
+        run: |
+          cilium install \
+            --chart-directory install/kubernetes/cilium \
+            --config monitor-aggregation=none \
+            --helm-set loadBalancer.l7.backend=envoy \
+            --helm-set tls.secretsBackend=k8s \
+            --agent-image=quay.io/cilium/cilium:test \
+            --operator-image=quay.io/cilium/operator-generic:test \
+            --helm-set image.pullPolicy=IfNotPresent \
+            --helm-set operator.image.pullPolicy=IfNotPresent \
```
sayboras, line 127: we can just change pull policy to never

.github/workflows/cilium-integration-tests.yaml
```diff
@@ -0,0 +1,155 @@
...
+        shell: bash
+        run: DOCKER_IMAGE_TAG=test make docker-cilium-image docker-operator-generic-image
+
+      - name: Load Cilium Images into kind
+        shell: bash
+        run: |
+          kind load docker-image \
+            --name kind \
+          quay.io/cilium/operator-generic:test \
+          quay.io/cilium/cilium:test
```
sayboras, line 114: nit: indentation alignment.

.github/workflows/cilium-integration-tests.yaml
```diff
@@ -0,0 +1,155 @@
...
+  CILIUM_CLI_TAG: latest
+
+jobs:
+  cilium-connectivity-tests:
+    timeout-minutes: 360
+    name: Cilium Connectivity Tests
+    runs-on: ubuntu-latest-64-cores-256gb
+    steps:
+      - name: Prepare variables for pushing to master
+        if: github.event_name == 'push'
```
sayboras, line 22: do we need to check for base ref here i.e. branch is master ?

.github/workflows/cilium-integration-tests.yaml
```diff
@@ -0,0 +1,155 @@
...
+            --agent-image=quay.io/cilium/cilium:test \
+            --operator-image=quay.io/cilium/operator-generic:test \
+            --helm-set image.pullPolicy=IfNotPresent \
+            --helm-set operator.image.pullPolicy=IfNotPresent \
+            --config debug=true \
+            --config debug-verbose=envoy
+
+      - name: Execute Cilium Connectivity Tests
+        shell: bash
+        run: cilium connectivity test
```
sayboras, line 133: it's open question, do we need to enable hubble for flow validation as part of the connectivity tests ?

sayboras (commented):
this is a huge improvement :+1:  :1st_place_medal:

.github/workflows/cilium-integration-tests.yaml
mhofstetter: good point, changed it to `/test-cilium-integration`

.github/workflows/cilium-integration-tests.yaml
mhofstetter: yes, because otherwise we might trigger the workflow twice when pushing to a branch with a corresponding PR (push & PR sync event)

.github/workflows/cilium-integration-tests.yaml
mhofstetter: yeah, it looks like the cilium connectivity tests check some additional stuff when hubble relay is available. thanks!

.github/workflows/cilium-integration-tests.yaml
sayboras: hehe i mean just /test :D

mhofstetter: oh :D then i tend to "no" because we're not triggering every test - which i would expect from a trigger comment `/test`. it's quite explicit that we're executing the cilium integration tests with explicit version of cilium and/or cilium cli.

:thinking:

.github/workflows/cilium-integration-tests.yaml
mhofstetter: @sayboras we might need to adapt this one once we branch new release branches from master

.github/workflows/cilium-integration-tests.yaml
```diff
@@ -0,0 +1,160 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Prepare variables for pushing to master
+        if: github.event_name == 'push' && github.ref == 'refs/heads/master'
+        run: |
+          echo "PROXY_IMAGE=quay.io/cilium/cilium-envoy" >> $GITHUB_ENV
+          echo "PROXY_TAG=${{ github.sha }}" >> $GITHUB_ENV
+
+      - name: Prepare variables for PR
+        if: github.event_name == 'pull_request'
```
mhofstetter, line 28: i realized that we need to filter out PR's (and comments on PR's) which aren't targeting `master`. as mentioned in another comment, we need to adapt once we're branching a new release branch from `master` and want to support it there too.

.github/workflows/cilium-integration-tests.yaml
mhofstetter: good point. added steps to collect and download system dump in case of failure

.github/workflows/cilium-integration-tests.yaml
mhofstetter: adapted, that PR's based cilium integration tests are only triggered if the base branch of the PR is `master`. on other PR's there's still the possibility to use the variant with the issue comment on the PR.

in the future we might chose the right version of cilium & cilium-cli for other proxy release branches as well.

.github/workflows/cilium-integration-tests.yaml
mhofstetter: actually it's also covered by the branch filter in the "on push" event definition. going to remove it.

mhofstetter:
@sayboras can you please have a second look?

i addressed your comments and some other points (version pinning, define conditions on the job itself (e.g. define event & job  conditions, such that the workflow only gets triggered on pushes to master, PRs targeting master and issue comments on PRs starting with the right prefix. IMO this then allows us to drop these tests on steps of this job)

sayboras (approved):
Thanks and LGTM :100:.

Just a note that we will need to backport this change to v1.23 branch later.

jrajahalme (approved):
Thanks for this, saves a ton of time automating this!

Some nits below.

.github/workflows/cilium-integration-tests.yaml
```diff
@@ -0,0 +1,187 @@
...
+          cid=$(docker create quay.io/cilium/cilium-cli-ci:${{ env.CILIUM_CLI_TAG }} ls)
+          sudo docker cp $cid:/usr/local/bin/cilium /usr/local/bin
+          docker rm $cid
+          cilium version
+
+      - name: Create kind cluster
+        uses: helm/kind-action@d8ccf8fb623ce1bb360ae2f45f323d9d5c5e9f00 # v1.5.0
+        with:
+          version: ${{ env.KIND_VERSION }}
+          config: '.github/kind-config.yaml'
```
jrajahalme, line 101: Use of a kind cluster with two nodes would provide more coverage.

.github/workflows/cilium-integration-tests.yaml
```diff
@@ -0,0 +1,187 @@
...
+            --config debug-verbose=envoy
+
+          cilium hubble enable \
+            --chart-directory install/kubernetes/cilium
+
+          cilium hubble port-forward&
+
+      - name: Execute Cilium Connectivity Tests
+        shell: bash
+        run: cilium connectivity test
```
jrajahalme, line 149: IMO we could limit the tests with `--test=l7` option, that would limit the test to those with `l7` in their name, which would be all we need for Envoy integration. This would shave couple of minutes off the test time.

jrajahalme (commented):
Some additional comments for future improvement :-)

jrajahalme:
Seems that run on PR push does not work, says the test is skipped: https://github.com/cilium/proxy/actions/runs/4624574857/jobs/8179587054?pr=142

(The PR in question was rebased to master, if it makes a difference)

.github/workflows/cilium-integration-tests.yaml
```diff
@@ -0,0 +1,187 @@
...
+        if: github.event_name == 'pull_request'
+        run: |
+          echo "PROXY_IMAGE=quay.io/cilium/cilium-envoy-dev" >> $GITHUB_ENV
+          echo "PROXY_TAG=${{ github.event.pull_request.head.sha }}" >> $GITHUB_ENV
+
+      - name: Prepare variables for issue comment on a PR
+        if: github.event_name == 'issue_comment'
+        run: |
+          echo "PROXY_IMAGE=quay.io/cilium/cilium-envoy-dev" >> $GITHUB_ENV
+          echo "PROXY_TAG=${{ github.event.issue.pull_request.head.sha }}" >> $GITHUB_ENV
```
jrajahalme, line 47: This fails on an issue comment:
```
  echo "PROXY_IMAGE=quay.io/cilium/cilium-envoy-dev" >> $GITHUB_ENV
  echo "PROXY_TAG=" >> $GITHUB_ENV
```
See the empty `PROXY_TAG`. As seen on: https://github.com/cilium/proxy/actions/runs/4624591204/jobs/8179582498

The `Wait for Cilium proxy image to be available` step never completes because of this.

mhofstetter: yeah, seems like `${{ github.event.issue.pull_request.head.sha }}` is not set in case of a comment :man_facepalming:

.github/workflows/cilium-integration-tests.yaml
```diff
@@ -0,0 +1,187 @@
...
+          commentBody="${{ github.event.comment.body }}"
+
+          ciliumRef=${CILIUM_REPO_REF}
+          if [[ "$commentBody" == *" cilium="* ]]; then
+            ciliumRef=$(echo "$commentBody" | sed -E 's|.* cilium=([^ ]*).*|\1|g')
+          fi
+          echo "CILIUM_REPO_REF=${ciliumRef}" >> $GITHUB_ENV
+
+          ciliumCliTag=${CILIUM_CLI_TAG}
+          if [[ "${{ github.event.comment.body }}" == *" ciliumCli="* ]]; then
```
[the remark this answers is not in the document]
mhofstetter: addressed in https://github.com/cilium/proxy/pull/159

.github/workflows/cilium-integration-tests.yaml
mhofstetter: addressed by https://github.com/cilium/proxy/pull/160

.github/workflows/cilium-integration-tests.yaml
```diff
@@ -0,0 +1,187 @@
...
+
+      - name: Checkout Cilium ${{ env.CILIUM_REPO_REF }}
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
+        with:
+          repository: cilium/cilium # Be aware that this is the Cilium repository and not the one of the proxy itself!
+          ref: ${{ env.CILIUM_REPO_REF }}
+
+      - name: Install Cilium CLI ${{ env.CILIUM_CLI_TAG }}
+        run: |
+          cid=$(docker create quay.io/cilium/cilium-cli-ci:${{ env.CILIUM_CLI_TAG }} ls)
```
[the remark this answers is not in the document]
mhofstetter: definitely. i switched from installing the Cilium CLI via `tar.gz` download from GitHub to the Docker Image approach because the `tar.gz` are only provided for releases whereas a Docker Image is built for every build.

But agree that it's not as comfortable to provide in case of a stable release. will address this together with the other mentioned improvements from you!

.github/workflows/cilium-integration-tests.yaml
```diff
@@ -0,0 +1,187 @@
...
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              comment_id: ${{ github.event.comment.id }},
+              content: 'rocket'
+            })
+
+      - name: Checkout Cilium ${{ env.CILIUM_REPO_REF }}
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
+        with:
+          repository: cilium/cilium # Be aware that this is the Cilium repository and not the one of the proxy itself!
```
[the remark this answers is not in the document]
mhofstetter: @jrajahalme don't get this point. 

this is just the repository. but one has the possibility to provide the `ref` when checking out the repo (https://github.com/cilium/proxy/blob/master/.github/workflows/cilium-integration-tests.yaml#L88) - which can be anything - including your feature branch name of the CIlium Repository.

.github/workflows/cilium-integration-tests.yaml
mhofstetter: @jrajahalme the config file already removes the taint from the master node to leverage both nodes.

https://github.com/cilium/cilium/blob/master/.github/kind-config.yaml#L6-L13

.github/workflows/cilium-integration-tests.yaml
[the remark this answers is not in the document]
mhofstetter: ah - forked repos aren't addressed - if that's what you mean.

.github/workflows/cilium-integration-tests.yaml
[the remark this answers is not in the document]
mhofstetter: addressed with PR https://github.com/cilium/proxy/pull/161

.github/workflows/cilium-integration-tests.yaml
mhofstetter: Addressed with https://github.com/cilium/proxy/pull/165

.github/workflows/cilium-integration-tests.yaml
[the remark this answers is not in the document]
mhofstetter: addressed with https://github.com/cilium/proxy/pull/166

mhofstetter:
@jrajahalme addressed all your input in PR's. thanks again!

PR #161, opened 2023-04-06
ci: cilium integration test supports stable version of cilium cli

mhofstetter:
Currently, the Cilium Integration test is installing Cilium CLI via Docker Image to support any given build of the CLI. The tar.gz only exist for actual CLI releases.

However, this comes with the drawback that referencing a stable release isn't that easy, because this needs knowledge of the actual GIT commit hash of the release.

Therefore, this commit installs the Cilium CLI via tar.gz when the Cilium CLI reference is a version (vx.x.x).


sayboras (approved):
LGTM :heavy_check_mark:

PR #164, opened 2023-04-09
envoy: Bump envoy version to v1.22.10

sayboras:
This is for latest patch release from upstream

https://github.com/envoyproxy/envoy/releases/tag/v1.22.10


sayboras:
Considering that CI jobs are all good, and we have applied and tested these security patches in the higher envoy version successfully, I am merging this in.

PR #165, opened 2023-04-11
ci: only execute L7 Connectivity tests when testing Cilium Integration

mhofstetter:
Currently all Cilium Connectivity tests are executed when testing the Cilium Integration.

In the context of the Cilium Proxy, only the L7 tests are actually interesting. Therefore, this commit adds the necessary flag `--test=l7`.

### DIFF
--- a/.github/workflows/cilium-integration-tests.yaml
+++ b/.github/workflows/cilium-integration-tests.yaml
@@ -153,9 +153,9 @@ jobs:
 
           cilium hubble port-forward&
 
-      - name: Execute Cilium Connectivity Tests
+      - name: Execute Cilium L7 Connectivity Tests
         shell: bash
-        run: cilium connectivity test
+        run: cilium connectivity test --test=l7
 
       - name: Reporting success to issue comment
         uses: actions/github-script@d7906e4ad0b1822421a7e6a35d5ca353c962f410 # v6.4.1


sayboras (approved):

.github/workflows/cilium-integration-tests.yaml
```diff
@@ -153,9 +153,9 @@ jobs:
 
           cilium hubble port-forward&
 
-      - name: Execute Cilium Connectivity Tests
+      - name: Execute Cilium L7 Connectivity Tests
         shell: bash
-        run: cilium connectivity test
+        run: cilium connectivity test --test=l7
```
sayboras, line 158: are the new TLS tests part of l7 suite? IMO, if it's not taking much longer time, we can just run all the connectivity tests as we might add tests without *l7* name in the future.

mhofstetter: @sayboras good point - thanks for the input. but i would rather ensure that we select the right test-suites via filter and/or have the tests named correctly (to match the filters) than just enabling all.

We would save quite some time (~5min) - and at the end it's not just about time but also to reduce false positives.

`client-egress-l7-tls-deny-without-headers` & `client-egress-l7-tls-headers` are named correctly (i assume you meant them with "new TLS tests?).

mhofstetter: btw: i assume that we don't cover everything with the connectivity tests. e.g. we could also execute `ingress` & `gateway api` conformence tests here. but i think at the end it's not the goal to cover everything in this repository.

PR #167, opened 2023-04-11
Support Ingress ID also for east/west Ingress

jrajahalme:
When a local pod accesses services through a Cilium Ingress, both the egress policy enforcement and ingress policy enforcement (at the destination) were performed based on the identity of the original source pod, in contrast to the case where an external user accesses these same backends via the same Cilium Ingress. In this case the ultimate destination does policy enforcement on the basis of the sender being the Ingress rather than the original source of the traffic.

Unify Ingress policy enforcement behaviour so that all traffic via Ingress is treated the same way, as "north/south" L7 LB traffic. This makes the upstream connection from Cilium Ingress to assume the Ingress identity for both the egress policy enforcement at the Ingress node (the source node if the source is a local pod), and in the destination node, where the source will be seen as Ingress also when the source is a local pod.

East/west L7 LB without Cilium Ingress is still supported, and in that case Ingress identity is not assumed.

These two modes of operation (L7 LB is east/west or north/south for local source pods) is controlled by the bpf_metadata filter configuration as follows, when `is_l7lb=true`:
- if `use_original_source_address` is `true`, the traffic is treated as local east/west traffic, no ingress identity is assumed
- if `use_original_source_address` is `false`, the traffic is treated as north/south even for local sources.

Cilium proxy does not have a special notion for Ingress identity, but resolves it via the ipcache when `ipv4_source_address` and/or `ipv6_source_address` options are configured. Cilium agent configures these for Ingress listeners.


jrajahalme:
Rebased to resolve conflicts

go/cilium/api/bpf_metadata.pb.go
```diff
@@ -36,13 +36,16 @@ type BpfMetadata struct {
...
-	// egress.
-	EgressMarkSourceEndpointId bool `protobuf:"varint,4,opt,name=egress_mark_source_endpoint_id,json=egressMarkSourceEndpointId,proto3" json:"egress_mark_source_endpoint_id,omitempty"`
+	UseOriginalSourceAddress bool `protobuf:"varint,3,opt,name=use_original_source_address,json=useOriginalSourceAddress,proto3" json:"use_original_source_address,omitempty"`
+	// True if the listener is used for an L7 LB. In this case policy enforcement is done on the
+	// destination selected by the listener rather than on the original destination address. For
+	// local sources the source endpoint ID is set in socket mark instead of source security ID if
+	// 'use_original_source_address' is also true, so that the local source's egress policy is
+	// enforced on the bpf datapath.
+	// Only valid for egress.
+	IsL7Lb bool `protobuf:"varint,4,opt,name=is_l7lb,json=isL7lb,proto3" json:"is_l7lb,omitempty"`
```
sayboras, line 48: with the rename here, can we still backport this change to the older releases (e.g. cilium 1.13 with envoy 1.23). Normally, we didn't update API in old releases.

sayboras (commented):
The high-level changes are making sense to me i.e. we are using one more indicator (e.g. use_original_source_address) to decide which identity will be used for policy enforcement. 

Just one comment on backport feasibility.

go/cilium/api/bpf_metadata.pb.go
jrajahalme: Old versions can use the old API definitions with old names.

jrajahalme: So the backport would be different in that regard.

jrajahalme: Or then we can also backport the name change, it's not really a big deal since we do not go older than Cilium 1.13.

tests/metadata_config_test.cc
```diff
@@ -0,0 +1,341 @@
...
+    socket_.connection_info_provider_ =
+        std::make_shared<Network::ConnectionInfoSetterImpl>(local_address_, remote_address_);
+
+    initTestMaps(context_);
+
+    Init::WatcherImpl watcher("metadata test", []() {});
+    context_.initManager().initialize(watcher);
+
+    config_ = std::make_shared<Cilium::BpfMetadata::Config>(config, context_);
+    filter_ = std::make_unique<Cilium::BpfMetadata::Instance>(config_);
```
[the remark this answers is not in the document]
jrajahalme: Thanks, it was a remnant from before I figured I can test this just with the config, removed.

go/cilium/api/bpf_metadata.pb.go
sayboras: yeah, i was thinking about backport this to 1.12 as well. But we can adapt the patch as well.

sayboras (approved):
LGTM :heavy_check_mark:

PR #168, opened 2023-04-13
ci: Use Cilium Version as docker image tag in Cilium Integration Tests

mhofstetter:
Currently the tag "test" is used when building the Cilium Agent & Operator Docker images during the Cilium Integration Tests.

This results in errors when the connectivity tests are trying to parse the tag as semantic version.

This commit uses the version from the VERSION file within the Cilium repository as version.

Even though we're fixing the version evaluation issue with https://github.com/cilium/cilium-cli/pull/1510 - it's probably a good idea to use a proper semver.


mhofstetter:
Successful run with `pull_request` event: https://github.com/cilium/proxy/actions/runs/4686842231/jobs/8305392464?pr=168

sayboras (approved):

PR #173, opened 2023-04-15
patches: Update for upstreamed versions

jrajahalme:
Replace the TCP proxy and router patches for the versions that are being upstreamed. This removes the added `receive_before_connect` configuration option from the TCP proxy filter; the corresponding filter state metadata must be used instead.

The earlier we apply the TCP Proxy API change the better. Even if there would be minor edits in the upstreaming process, these patches can be removed once we update to a version of Envoy that has these upstream changes.

Remove the tcmalloc patch, as the version used by Envoy now works also for Arm builds.


sayboras (approved):
Nice :heavy_check_mark:, 1 down and 4 (and a half) to go :heavy_check_mark:

PR #185, opened 2023-04-21
admin: include cilium networkpolicies in envoy config dump

sayboras:
[usptream commit 797bea843de17da8f8a096747c4691405b540aa8]

Currently, only builtin Envoy resources are part of the Envoy config dump on the admin API (/admin/config_dump).

This commit adds Ciliums NetworkPolicies to the config dump, by registering a callback to the configtracker of the adminserver.

This way, the network policies which are actually taken into account by the Envoy Cilium filters are also part of Ciliums sysdump.


sayboras:
There are some conflicts as admin() context is a pointer instead of an optional ref in v1.23.

go/cilium/api/npds.pb.go
mhofstetter, line 1: i get a diff on the generated proto go file when i execute `make api` locally on your branch. have you freshly generated them? (i assume this is necessary)

sayboras: ah good point, i thought there is a check in gha :face_exhaling:

sayboras: done and thanks

mhofstetter (approved):
LGTM, thanks for the backport!

PR #187, opened 2023-04-26
image: use ubuntu-22 as base image instead of scratch

mhofstetter:
Currently, the docker image quay.io/cilium/cilium-envoy builds on top of plain scratch and only contains the cilium-envoy binary. It's only used for "transport" purposes.

With the introduction of a dedicated DaemonSet for Envoy in the context of Cilium, it becomes necessary to add a "runtime" to the cilium envoy docker image. This commit starts off by using Ubuntu. In a next step, we might try to use a smaller base image.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -7,6 +7,7 @@
   ],
   "includePaths": [
     ".github/workflows/**",
+    "Dockerfile",
     "Dockerfile.builder",
     "WORKSPACE",
     "ENVOY_VERSION"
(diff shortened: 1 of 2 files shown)


sayboras (approved):
LGTM.

With this change, we even can run the built image directly (instead of copying binary to host) in the envoy version check step.

```
$ docker run -it --rm quay.io/cilium/cilium-envoy-dev:b6c8a0e8aa0a07b22cd90e34d9547041bf7000dd@sha256:56bbfb690e75081c5da040f32cb7bdebde30152d8a17d22a74cc301b1b79d1ca cilium-envoy --version

cilium-envoy  version: b6c8a0e8aa0a07b22cd90e34d9547041bf7000dd/1.25.5/Distribution/RELEASE/BoringSSL
```

PR #205, opened 2023-04-28
ci: fix envoy version check issue

mhofstetter:
This commit removes the interactive flag when executing the envoy version check to prevent the error `The input device is not a TTY`.

fixes: https://github.com/cilium/proxy/pull/190


mhofstetter:
Build now fails with this error due to the fact that this workflow uses `pull_request_target` - which was the reason CI didn't detect it on the initial PR

PR #212, opened 2023-05-15
ci: sign container image with cosign

mhofstetter:
This PR implements container image signing using cosign.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -132,6 +132,17 @@
     },
     {
       "fileMatch": [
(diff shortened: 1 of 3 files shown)


mhofstetter:
Successful run of build with `pull_request` instead of `pull_request_target`: https://github.com/cilium/proxy/actions/runs/4979997112/jobs/8912321483?pr=212

sayboras (approved):
CI image signing might not be required, but we can revisit again depending on how flaky this signing process is. LGTM :heavy_check_mark:

PR #216, opened 2023-05-31
Fix policy update crash

jrajahalme:
Recently introduced SDS support for secrets in Cilium network policies causes an Envoy crash no assert failure due to a resource being freed on a different thread than where it was allocated. Fix this by processing policy updates on the main thread after all the worker threads have been updated. This was the last reference to an old policy is freed on the main thread, where all the policy objects are allocated.

Also limit the SDS initial fetch timeout to 5 seconds from the default 15 seconds.


mhofstetter (approved):
LGTM - although it might be useful for someone with a bit more C++ knowledge to take a look at it too.

nit: it seems that we're mixing tabchar & spaces for indention quite heavily in this codebase :smile:

PR #218, opened 2023-06-01
Dockerfile: Install ca-certificates package

jrajahalme:
CA certificates may be needed for certificate validation. When they are installed, Envoy can be told to use the CA certs in the file system by creating a file-based Envoy secret:
```
  secrets:
  - name: validation_context
    validation_context:
      trusted_ca:
        filename: /etc/ssl/certs/ca-certificates.crt
```

This can then be referred to in TLS config:
```
  transport_socket:
    name: envoy.transport_sockets.tls
    typed_config:
      "@type": type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.UpstreamTlsContext
      common_tls_context:
        validation_context_sds_secret_config:
          name: validation_context
```


mhofstetter (approved):
Good catch! 

I assume this have been missed during a test with Envoy as separate DaemonSet? Iwonder which other packages provided by the agent image (runtime image) might be missed in the Envoy DaemonSet deployment :thinking:

PR #221, opened 2023-06-02
build: sync extensions, enable internal listener

jrajahalme:
Sync extensions with both our v1.23 and upstream release/v1.25.
- enable all extensions that were already enabled in v1.23

Enable internal listener extensions and additional HTTP URI/header match/rewrite new in v1.25 (vs. v1.23).


mhofstetter (approved):

sayboras (approved):
:heavy_check_mark:

PR #266, opened 2023-07-07
Proxy id v1.24

jrajahalme:
Add proxy id field to both bpf_metadata filter config and access log messages. Access log messages will then contain the value configured on the bpf_metadata. This can be used to correlate access log messages to listeners explicitly by using the proxy port as the proxy id, for example.


jrajahalme:
Added updated generated API files.

PR #267, opened 2023-07-07
add proxy_id to bpf_metadata config and accesslog messages

jrajahalme:
Add proxy id field to both bpf_metadata filter config and acccess log messages. Access log messages will then contain the value configured on the bpf_metadata. This can be used to correlate access log messages to listeners explicitly by using the proxy port as the proxy id, for example.


sayboras (approved):
Just a note for future proxy upgrade for old releases that this requires api update in cilium/cilium

PR #268, opened 2023-07-07
bazel: Retry flaky tests

jrajahalme:
We re-trigger GH actions on flaky tests anyway, so let bazel test retry them automatically. Optimally we would not have any test flakes, but in practice this seems beneficial.


mhofstetter (approved):
like it - thanks!

sayboras (approved):

PR #271, opened 2023-07-13
build: Fix the main Envoy image tag name

haiyuewa:
Fix the typo 'DOCKER_IMAGE_ENVOY_TAGS', which will also create image with tag 'latest-'{arch} for main branch.


mhofstetter (approved):
Thanks for the contribution and fixing this issue @haiyuewa!

Be aware, that this is only relevant for locally built docker images.

We're already building and pushing multi-arch images on tag `latest` for `amd64` and `arm64` during CI: https://github.com/cilium/proxy/blob/main/.github/workflows/build-envoy-images-release.yaml#L202

PR #297, opened 2023-08-14
Pr/jrajahalme/drop privileges

jrajahalme:
Add a privileged thread early in the process bootstrap and drop all privileges from all other threads (e.g., Envoy main and worker threads). Call into the privileged thread to make syscalls for bpf map lookups and privileged socket options.


jrajahalme:
Merged in another PR.

PR #299, opened 2023-08-15
build: Add libcap-dev also for tests and vagrant

jrajahalme:
Add libcap-dev also in Dockerfile.builder.tests and Vagrantfile.

libcap is needed for the <sys/capability.h> header, the library is not linked in,
as there are no cross-compiled packages for it.


sayboras (approved):
✅

PR #302, opened 2023-08-15
ipcache: Sync map value definition with cilium main

jrajahalme:
Sync ipcache map value definition with cilium main. Only the first member is used, so this should have no functional difference.


sayboras (commented):
LGTM, the failure in integration is known flake

PR #303, opened 2023-08-15
build: Use apt-add-repository

jrajahalme:
Use apt-add-repository instead of directly editing apt sources files, as this is less error prone.

Without this I had the makefile adding multiple lines in a fresh VM where llvm install got into a
weird state. With this change the llvm source will appear at most once in the sources list.


sayboras (approved):
Seems like llvm15 repo is just added in focal apt ✅

PR #315, opened 2023-08-31
Use `cilium-envoy-starter` to run Envoy without privileges

jrajahalme:
Introduce a `cilium-envoy-starter` that forks and execs `cilium-envoy`
after having dropped all privileges.

Perform bpf map and privileged socket option operations in
`cilium-envoy-starter`, as requested by Cilium filters running in
`cilium-envoy` via a pipe between the two. Currently the protocol over
this pipe supports only the following operations:

- dump capabilities cilium-envoy-starter is running with
- open bpf maps
- perform bpf map lookups
- set socket options

cilium-envoy-starter fails to start if it does not have adequate
privileges needed for the above operations.

cilium-envoy now exits if it is running with any privileges when Cilium
filters are first configured.

Implementation detail:

- `libcap` is not used, as it would be hard to cross-compile for both arm64
  and amd64. Fortunately we only need to drop capabilities, so using the
  system call interface is not complicated.

- Patch Envoy to support setting socket options from listener filters and
  use them to set privileged options for the listener.

- Remove the unused support for creating and manipulating bpf maps.

- Remove `setrlimit` call that is only needed if the process is creating
  bpf maps.

- Minor cleanup in socket option implementation to move unused fields
  from SocketMarkOption class to SocketOption where they are used.

### DIFF
--- a/cilium/bpf.cc
+++ b/cilium/bpf.cc
@@ -1,16 +1,5 @@
 #include "cilium/bpf.h"
-
-#include <string.h>
-#include <sys/resource.h>
-#include <unistd.h>
-
-#include <cstdint>
-#include <fstream>
-#include <iostream>
-#include <map>
-#include <sstream>
-#include <string>
-#include <vector>
+#include "cilium/privileged_service_client.h"
 
 #include "source/common/common/utility.h"
 
@@ -24,10 +13,7 @@ enum {
 };
 
 Bpf::Bpf(uint32_t map_type, uint32_t key_size, uint32_t value_size)
-    : fd_(-1), map_type_(map_type), key_size_(key_size), value_size_(value_size) {
-  struct rlimit rl = {RLIM_INFINITY, RLIM_INFINITY};
-  setrlimit(RLIMIT_MEMLOCK, &rl);
-}
+    : fd_(-1), map_type_(map_type), key_size_(key_size), value_size_(value_size) {}
 
 Bpf::~Bpf() { close(); }
 
@@ -39,10 +25,10 @@ void Bpf::close() {
 
 bool Bpf::open(const std::string& path) {
   bool log_on_error = ENVOY_LOG_CHECK_LEVEL(trace);
-  union bpf_attr attr = {};
-  attr.pathname = uintptr_t(path.c_str());
 
-  fd_ = bpfSyscall(BPF_OBJ_GET, &attr);
+  auto& cilium_calls = PrivilegedService::Singleton::get();
+  auto ret = cilium_calls.bpf_open(path.c_str());
+  fd_ = ret.return_value_;
   if (fd_ >= 0) {
     // Open fdinfo to check the map type and key and value size.
     std::string line;
@@ -98,79 +84,20 @@ bool Bpf::open(const std::string& path) {
                 bpf_file_path);
     }
     close();
-  } else if (errno == ENOENT && log_on_error) {
+  } else if (ret.errno_ == ENOENT && log_on_error) {
     ENVOY_LOG(debug, "cilium.bpf_metadata: bpf syscall for map {} failed: {}", path,
-              Envoy::errorDetails(errno));
+              Envoy::errorDetails(ret.errno_));
   } else if (log_on_error) {
     ENVOY_LOG(warn, "cilium.bpf_metadata: bpf syscall for map {} failed: {}", path,
-              Envoy::errorDetails(errno));
+              Envoy::errorDetails(ret.errno_));
   }
 
   return false;
 }
 
-bool Bpf::create(uint32_t max_entries, uint32_t flags) {
-  union bpf_attr attr = {};
-  attr.map_type = map_type_;
-  attr.key_size = key_size_;
-  attr.value_size = value_size_;
-  attr.max_entries = max_entries;
-  attr.map_flags = flags;
-
-  fd_ = bpfSyscall(BPF_MAP_CREATE, &attr);
-  return fd_ >= 0;
-}
-
-bool Bpf::pin(const std::string& path) {
-  union bpf_attr attr = {};
-  attr.pathname = uintptr_t(path.c_str());
-  attr.bpf_fd = uint32_t(fd_);
-
-  return bpfSyscall(BPF_OBJ_PIN, &attr) == 0;
-}
-
-bool Bpf::insert(const void* key, const void* value) {
-  union bpf_attr attr = {};
-  attr.map_fd = uint32_t(fd_);
-  attr.key = uintptr_t(key);
-  attr.value = uintptr_t(value);
-  attr.flags = BPF_ANY;
-
-  return bpfSyscall(BPF_MAP_UPDATE_ELEM, &attr) == 0;
-}
-
-bool Bpf::remove(const void* key) {
-  union bpf_attr attr = {};
-  attr.map_fd = uint32_t(fd_);
-  attr.key = uintptr_t(key);
-  attr.flags = BPF_ANY;
-
-  return bpfSyscall(BPF_MAP_DELETE_ELEM, &attr) == 0;
-}
-
 bool Bpf::lookup(const void* key, void* value) {
-  union bpf_attr attr = {};
-  attr.map_fd = uint32_t(fd_);
-  attr.key = uintptr_t(key);
-  attr.value = uintptr_t(value);
-
-  return bpfSyscall(BPF_MAP_LOOKUP_ELEM, &attr) == 0;
-}
-
-#ifndef __NR_bpf
-#if defined(__i386__)
-#define __NR_bpf 357
-#elif defined(__x86_64__)
-#define __NR_bpf 321
-#elif defined(__aarch64__)
-#define __NR_bpf 280
-#else
-#error __NR_bpf not defined.
-#endif
-#endif
-
-int Bpf::bpfSyscall(int cmd, union bpf_attr* attr) {
-  return ::syscall(__NR_bpf, cmd, attr, sizeof(*attr));
+  auto& cilium_calls = PrivilegedService::Singleton::get();
+  return cilium_calls.bpf_lookup(fd_, key, key_size_, value, value_size_).return_value_ == 0;
 }
 
 } // namespace Cilium
--- a/starter/main.cc
+++ b/starter/main.cc
@@ -1,0 +1,108 @@
+#if !defined(__linux__)
+#error "Linux platform file is part of non-Linux build."
+#endif
+
+#include <errno.h>
+#include <syscall.h>
+#include <unistd.h>
+
+#include <sys/prctl.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+
+#include "starter/privileged_service_server.h"
+
+// NOLINT(namespace-envoy)
+
+#define STARTER_SUFFIX "-starter"
+#define STARTER_SUFFIX_LEN (sizeof(STARTER_SUFFIX) - 1)
+
+int main(int /* argc */, char** argv) {
+  // Check that we have the required capabilities
+  uint64_t caps = Envoy::Cilium::PrivilegedService::get_capabilities(CAP_EFFECTIVE);
+  if ((caps & (1UL << CAP_NET_ADMIN)) == 0 ||
+      (caps & (1UL << CAP_SYS_ADMIN | 1UL << CAP_BPF)) == 0) {
+    fprintf(stderr, "CAP_NET_ADMIN and either CAP_SYS_ADMIN or CAP_BPF capabilities are needed for Cilium datapath integration.\n");
+    exit(1);
+  }
+
+  // Get the path we're running from
+  char *path = new char[PATH_MAX];
+  constexpr size_t path_size = PATH_MAX;
+  int path_len = readlink("/proc/self/exe", path, path_size);
+  if (path_len < 0 || path_len >= int(path_size)) {
+    fprintf(stderr, "could not get path of the current executable: %s\n", strerror(errno));
+    exit(1);
+  }
+
+  // Remove the trailing "-starter" suffix.
+  // Check first that the executable name ends in the suffix
+  // and is not just the suffix.
+  if (size_t(path_len) > STARTER_SUFFIX_LEN && // more than suffix in path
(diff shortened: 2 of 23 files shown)


jrajahalme:
Cilium Integration test fails expectedly as Cilium Agent update is also needed so that it will launch Envoy via the new `cilium-envoy-wrapper`. Could also consider reversing the names so that `cilium-envoy` would be the wrapper and actual Envoy binary would be e.g., `cilium-envoy-wrapped`, but IMO it is clearer the way it is now in this PR.

jrajahalme:
> Cilium Integration test fails expectedly as Cilium Agent update is also needed so that it will launch Envoy via the new cilium-envoy-wrapper. Could also consider reversing the names so that cilium-envoy would be the wrapper and actual Envoy binary would be e.g., cilium-envoy-wrapped, but IMO it is clearer the way it is now in this PR.

Cilium Agent update is anyway needed to drop the socket options and `Transparent=true` from Listener config, so there is no way we can do this without agent update. Might as well go with the cleaner executable names.

cilium/envoy_wrapper_main.cc
```diff
@@ -0,0 +1,259 @@
...
+}
+
+} // namespace CiliumEnvoyWrapper
+
+// NOLINT(namespace-envoy)
+
+#define WRAPPER_SUFFIX "-wrapper"
+#define WRAPPER_SUFFIX_LEN (sizeof(WRAPPER_SUFFIX) - 1)
+
+int main(int /* argc */, char** argv) {
```
mhofstetter, line 171: after having a first look at the overall structure (new bazel libs and the `main` of the wrapper) i wonder whether it wouldn't be better to even further separate the wrapper logic from the "privileged service" bits.

Currently:
```mermaid
graph LR;
    W[Wrapper]-- forks -->E[Envoy];
    E-. calls .-> W;
```

Idea:
```mermaid
graph LR;
    W[Wrapper]-- forks -->E[Envoy];
    W[Wrapper]-- forks -->P[Privileged Service];
    E-. calls .-> P;
```

IMO this could lead to better understandable architecture and names (bazel libraries, C++ types, etc). Currently everything is named ~ `EnvoyWrapper**` which is kind of unclear.

* Wrapper: Being responsible to start the necessary processes. Maybe we could even rename this to "EnvoyStarter" to make this more clear.
* Privileged Service: Responsible to provide the privileged actions (BPF map access, socket handling) to the Envoy process. (Same here, we might find a more precise name)

And it might even support cases where the two processes (envoy & "privileged service") are started independent of the wrapper (starter) - e.g. by running the privileged service in a separate sidecar. Or at least we're preparing for this from an architectural point of view.

mhofstetter (commented):
Nice - already looks great! :rocket: 

I placed an "architectural" question (with a proposal) about the wrapper inline. Might be worth to discuss this first before I dive deeper into the details.

cilium/envoy_wrapper_main.cc
jrajahalme: > And it might even support cases where the two processes (envoy & "privileged service") are started independent of the wrapper (starter) - e.g. by running the privileged service in a separate sidecar. Or at least we're preparing for this from an architectural point of view.

This will need a different, more explicit method of passing the fd to the cilium-envoy process. I looked into this briefly, but did not test anything out yet, IMO we can leave this for now. However, if we implement support for the sidecar then the wrapper/starter is not necessary at all. So IMO building more structure into this now would be an overkill.

Better names for the bazel bits are welcomed, I'm pretty bad at naming these kind of things :-)

jrajahalme:
> Nice - already looks great! 🚀
> 
> I placed an "architectural" question (with a proposal) about the wrapper inline. Might be worth to discuss this first before I dive deeper into the details.

Detail would not change much if the wrapper would fork another process to serve the privileged requests, basically one more fork and the parent process waiting for this new process to finish - I think this is unnecessary, however, and would not help architecturally either.

So please delve into the detail :-)

cilium/envoy_wrapper_main.cc
mhofstetter: > > And it might even support cases where the two processes (envoy & "privileged service") are started independent of the wrapper (starter) - e.g. by running the privileged service in a separate sidecar. Or at least we're preparing for this from an architectural point of view.
> 
> This will need a different, more explicit method of passing the fd to the cilium-envoy process. I looked into this briefly, but did not test anything out yet, IMO we can leave this for now. However, if we implement support for the sidecar then the wrapper/starter is not necessary at all. So IMO building more structure into this now would be an overkill.
> 
> Better names for the bazel bits are welcomed, I'm pretty bad at naming these kind of things :-)

I agree that the sidecar idea is a bit off for the time being. This shouldn't be the reason.

But as mentioned - it would also help in better understanding the design by separating the responsibilities - and therefore allow for better naming. Naming is especially hard if "the thing" is responsible for multiple tasks. But we might can solve this by splitting up `envoy_wrapper_main.cc` into these two aspects.

mhofstetter: I'd recommend to separate the main function (which is the actual wrapper / starter) from the "ProtocolServer by moving the ProtocolServer bits into its own file `envoy_wrapper_protocolserver.cc`.

And renaming all `envoy_wrapper_*`  (except `envoy_wrapper_main.cc`)  files to something `privileged_service_*` 

-> 

* `privileged_service_client`
* `privileged_service_server`
* `privileged_service_protocol`

and adapt the bazel envoy libs (and their names accordingly)

cilium/envoy_wrapper_main.cc
```diff
@@ -0,0 +1,259 @@
...
+  if (pid == 0) {
+    // in child process, close the parent end of the pipe
+    close(fds[0]);
+
+    // Unconditionally drop all capabilities
+    struct __user_cap_header_struct hdr{_LINUX_CAPABILITY_VERSION_3, 0};
+    struct __user_cap_data_struct data[2];
+    memset(&data, 0, sizeof(data));
+    if (::syscall(SYS_capset, &hdr, &data, sizeof(data)) != 0) {
+      perror("capset");
```
mhofstetter, line 219: Provide more meaningful error messages in all the `perror` calls

cilium/envoy_wrapper_main.cc
jrajahalme: Renamed and separated as suggested. Also renamed the new binary as `cilium-envoy-starter`, but kept it in a single process that now implements the `PrivilegedServiceServer`.

cilium/envoy_wrapper_main.cc
mhofstetter: sounds reasonable!

starter/privileged_service_client.h
```diff
@@ -0,0 +1,53 @@
...
+public:
+  ProtocolClient();
+
+  // allow access to the classes that need it
+  friend class Envoy::Cilium::Bpf;
+  friend class Envoy::Cilium::SocketMarkOption;
+
+protected:
+  // Read-only bpf syscalls
+  Envoy::Api::SysCallIntResult bpf_open(const char *path);
```
mhofstetter, line 31: document the 3 "public" facing functions?

starter/privileged_service_client.h
```diff
@@ -0,0 +1,53 @@
...
+  Envoy::Api::SysCallIntResult bpf_lookup(int fd, const void *key, uint32_t key_size, void* value,
+					  uint32_t value_size);
+
+  // Set a socket option
+  Envoy::Api::SysCallIntResult setsockopt(int sockfd, int level, int optname, const void *optval,
+					  socklen_t optlen);
+
+private:
+  bool check_privileged_service();
+  bool running_with_cilium_privileged_service() const { return is_open(); }
```
mhofstetter, line 41: rename to `connected_to_cilium_privileged_service` ?

starter/privileged_service_server.cc
```diff
@@ -0,0 +1,146 @@
...
+		     msg.setsockopt_req.optname_, &msg.setsockopt_req.optval_,
+		     sizeof(msg.setsockopt_req.optval_));
+      break;
+    default:
+      fprintf(stderr, "Unexpected privileged call type: %d\n", msg.hdr.msg_type_);
+      rc = -1;
+      errno = EINVAL;
+    }
+
+    // Close the received file descriptor
```
mhofstetter, line 118: nit: `receiving file descriptor`?

starter/privileged_service_server.cc
```diff
@@ -0,0 +1,146 @@
...
+      msg.response.ret_ = {INT_MAX, 0};
+    } else {
+      msg.response.ret_ = {rc, rc != -1 ? 0 : errno};
+    }
+    size = send_fd_msg(&msg, sizeof(msg.response), buf, value_len, fd_out);
+    if (size < ssize_t(sizeof(msg.response) + value_len)) {
+      perror("sendmsg");
+    }
+
+    // Close the sent file descriptor
```
mhofstetter, line 137: nit: `sending file descriptor`?

cilium/envoy_wrapper_main.cc
mhofstetter: nice - looks great! thanks!

mhofstetter (changes requested):
Nice improvement by splitting out the privileged service (&starter)! :rocket: 

Some suggestions from the next review iteration inline.

cilium/bpf.cc
```diff
@@ -39,10 +25,10 @@ void Bpf::close() {
 
 bool Bpf::open(const std::string& path) {
   bool log_on_error = ENVOY_LOG_CHECK_LEVEL(trace);
-  union bpf_attr attr = {};
-  attr.pathname = uintptr_t(path.c_str());
 
-  fd_ = bpfSyscall(BPF_OBJ_GET, &attr);
+  auto& cilium_calls = PrivilegedService::Singleton::get();
```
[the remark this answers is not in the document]
jrajahalme: Not sure, I kind of like to keep this name Cilium specific (see other comments below).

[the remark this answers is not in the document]
mhofstetter: `cilium_privileged_service`?

starter/privileged_service_client.h
jrajahalme: I'm thinking this privileged API will morph to something more CIlium specific in near future. Giving access to set any/all socket options and open any bpf/all bpf maps is not the most secure thing to do. Both the bpf lookup and socket option setting happen via the bpf_metadata filter, so there would be an opportunity to lift the abstraction level of this privileged API from the current syscall level to something like "Cilium connection metadata setup" that would:
- Get the metadata for the new connection from conntrack and/or ipcache maps, and, if successful
- Set the socket options Cilium needs for new proxied connections
A separate call would be needed to set up the listening socket (socket options only)

This is a bit bigger change, shifting all the bpf code to the starter and refactoring the bpf_metadata filter code, so I did not want to start with this "bigger picture".

So in summary, I'd like to skip documenting this version of the privileged API, and would like to not characterise this as "public facing", even though it kind of is.

starter/main.cc
```diff
@@ -0,0 +1,104 @@
...
+    if (fds[1] != CILIUM_PRIVILEGED_SERVICE_FD) {
+      if (dup2(fds[1], CILIUM_PRIVILEGED_SERVICE_FD) < 0) {
+	perror("dup2");
+	exit(1);
+      }
+      close(fds[1]);
+    }
+    
+    // Exec cilium-envoy process
+    execv(path, argv);
```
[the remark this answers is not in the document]
jrajahalme: I've tried. Currently the server side exits if the client side closes the socket, and the destructor you mention waits for the client process to exit. Client side closes the socket on destruction, but maybe this never happens as Envoy:: ThreadSafeSingleton seems to not ever destruct.

However, the client side of the socket will get closed when the process exits, so the starter should gracefully exit right after, and the wait for the client process should be a no-op as it has already exited.

starter/privileged_service_server.cc
jrajahalme: No, it is literally closing the file descriptor received on the control channel of `recvmsg`.

starter/privileged_service_server.cc
jrajahalme: Same here, when opening a bpf map, the opened file descriptor is sent in recvmsg control channel and we need to close the sent file descriptor, as otherwise the bpf map would stay open due to the server side keeping an open file descriptor to it.

starter/main.cc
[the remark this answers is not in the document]
mhofstetter: great, thanks for the explanation!

starter/privileged_service_client.h
mhofstetter: makes sense. thanks for the outlook and explanation!

starter/privileged_service_client.h
jrajahalme: IMO this was (already) getting too long, renamed as `have_cilium_privileged_service()`

starter/privileged_service_client.cc
```diff
@@ -0,0 +1,131 @@
...
+  DumpRequest req;
+  Response resp;
+  uint8_t buf[1024];
+  int fd = -1;
+  ssize_t size = transact(req.hdr_, sizeof(req), nullptr, 0, &fd, resp, buf, sizeof(buf), false);
+  if (size < ssize_t(sizeof(resp))) {
+    return false;
+  }
+  std::string str(reinterpret_cast<char *>(buf), size - sizeof(resp));
+  ENVOY_LOG_MISC(debug, "Running with Cilium privileged service with the following capabilities: {}", str);
```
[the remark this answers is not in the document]
jrajahalme: rephrased as: `Cilium privileged service detected with following capabilities: {}`

mhofstetter (approved):
LGTM :rocket:

jrajahalme:
Cilium connectivity tests fail expectedly due to Envoy having been started with privileges without `cilium-envoy-starter`:
```
2023-09-11T17:18:27.392970338Z level=error msg="[assert failure: get_capabilities(CAP_EFFECTIVE) == 0 && get_capabilities(CAP_PERMITTED) == 0. Details: cilium-envoy running with privileges, exiting" subsys=envoy-assert threadID=979
```
Merging to main and posting an update on `cilium/cilium` to pick this up :-)

PR #353, opened 2023-09-13
cilium integration tests: increase "wait-for-images" timeout

mhofstetter:
Cilium Integration Tests often hit the timeout of 30m waiting for the proxy images being available.

This commit increases the wait timeout to 45m.

-> https://github.com/cilium/proxy/commits/main ~ more or less all runs on `main`

### DIFF
--- a/.github/workflows/cilium-integration-tests.yaml
+++ b/.github/workflows/cilium-integration-tests.yaml
@@ -171,7 +171,7 @@ jobs:
           go mod vendor
 
       - name: Wait for Cilium Proxy image to be available
-        timeout-minutes: 30
+        timeout-minutes: 45
         shell: bash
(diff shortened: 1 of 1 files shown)


jrajahalme:
Build should not take so long, but I guess normally increasing the timeout should not hurt.

We has a permissions problem with the bazel archive after we moved to non-root builds. It was fixed by removing the old archive and having one built from scratch. It is possible that an update path will mess with the permissions, though.

jrajahalme:
Did you see why the build took so long? Was it queued for a long time, or did it have a problem with the bazel archive?

mhofstetter:
> Did you see why the build took so long? Was it queued for a long time, or did it have a problem with the bazel archive?

in most of the cases the image build job was queued up whereas the integration test was already running.

jrajahalme (approved):

PR #373, opened 2023-10-04
Support deny rules and overlapping port ranges in policy enforcement

jrajahalme:
Add support for deny policies and overlapping port ranges in policy enforcement. This is needed for:
- policy correctness when enforcing ingress policy for Cilium Ingress
- allow Cilium Network Policies with named ports that resolve to the same port number used elsewhere in the policy (a trivial kind of overlapping port ranges).
- for when Cilium starts supporting port ranges in policies at a later time.

Cilium Agent needs to be updated to properly send deny policies to cilium-envoy.

Notes to reviewers:
- Review by commit: I separated refactoring to separate commits to make the rest more readable


sayboras:
Seems like there are a couple of conflicts need to be resolved.

jrajahalme:
Rebased and added support for overlapping port ranges, which also addresses policy having the same port multiple times, maybe due to named ports being used in Cilium Network Policy.

jrajahalme:
Added more test validation for range split correctness.

PR #379, opened 2023-10-10
Fix format

jrajahalme:
Add `starter` and `vendor` to the list of excluded directories and apply `make fix`.

This helps make future PRs cleaner, and allows to add a CI step for format check.


sayboras (approved):
Ah I forgot about these two targets (e.g. fix and check).

jrajahalme:
Pushed small fix to remove `Makeflie.dev` from `Dockerfile.tests.dockerignore`.

PR #385, opened 2023-10-11
v1.25: Fix archive tags to not collide with main

jrajahalme:
Use the release branch name instead of the main in archive tags, like in v1.24.


sayboras (approved):

sayboras (approved):
Thanks

PR #418, opened 2023-10-23
docs: Add UPGRADE.md step

sayboras:


UPGRADE.md
```diff
@@ -0,0 +1,67 @@
+# Envoy Upgrade
```
mhofstetter, line 1: name the file `UPGRADE_ENVOY.md`?

UPGRADE.md
```diff
@@ -0,0 +1,67 @@
...
+### Sync up Bazel version
+
+New Envoy minor version might require new Bazel version.
+
+1. Update `.bazelversion` file.
+2. Sync up `WORKSPACE` file with upstream.
+
+```shell
+# Building a new builder image locally with your own docker account
+$ DOCKER_DEV_ACCOUNT=docker.io/sayboras ARCH=multi NO_CACHE=1 make docker-image-builder
```
mhofstetter, line 22: worth a note that this gets actually pushed to the registry due to not being able to use the local docker registry with multi arch builds? (docker CLI needs to be authenticated, image will be publicly accessible, ...)

same for actually building envoy docker image on line 59

mhofstetter (approved):
thanks @sayboras

UPGRADE.md
```diff
@@ -0,0 +1,67 @@
+# Envoy Upgrade
+
+Occasionally, we need to bump Envoy minor release version to support new
+upstream features, or for any security fixes with patch version.
+
+The recent PR [#417](https://github.com/cilium/proxy/pull/417) can be used
+as reference.
+
+For the patch release, we normally just need to do step (2) most of the time.
```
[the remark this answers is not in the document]
sayboras: ah right, good catch. Let me add the reference instead.

UPGRADE.md
sayboras: yup, that is correct.

Issue #308, opened 2023-08-23
Replace proxylib builder image


sayboras:
### Description

Currently, the proxylib builder image is from upstream quay.io/cilium/cilium-builder image, which can cause some problem for auto upgrade. 

It will be better to change to something else, which is smaller and also managed by renovate bot.

https://github.com/cilium/proxy/blob/f4eb004fd81e38ad05d4014fd5f5ff8e9bbf6fe8/Dockerfile#L6-L8

PR #420, opened 2023-10-26
docker: Use envoy builder image as proxylib builder

sayboras:
Fixes: https://github.com/cilium/proxy/issues/308

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,10 +3,6 @@
 #
 ARG BUILDER_BASE=quay.io/cilium/cilium-envoy-builder:6.1.0-latest
 
-# Common Builder image used in cilium/cilium
-# We need gcc for cgo cross-compilation at least, we can swap to something smaller later on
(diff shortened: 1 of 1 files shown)


.github/renovate.json5
```diff
@@ -60,6 +60,18 @@
         "v1.26",
       ]
     },
+    {
```
mhofstetter, line 63: i'd assume that we need an explicit regexMatcher that matches the image in the `ARG`.

By default [Renovate's language support for Docker](https://docs.renovatebot.com/docker/#how-it-works) only searches in `FROM` directives by default.

jrajahalme (commented):
It would be simpler to just use same cilium-envoy-builder image we already use in the build? It has Go and all the cross tooling as well.

sayboras:
> It would be simpler to just use same cilium-envoy-builder image we already use in the build? It has Go and all the cross tooling as well.

Ah good point, I forget that we recently add golang into builder image.

mhofstetter (approved):

PR #430, opened 2023-11-11
docs: Update version matrix

sayboras:
Also, as all the upstream versions are using v1.26+, we can also retire v1.25 branch.


mhofstetter (approved):
nice :partying_face: thanks!

PR #439, opened 2023-11-26
docker: Use envoy builder image as proxylib builder

sayboras:
Please refer to individual commit for more detail


mhofstetter (approved):
Thanks!

PR #440, opened 2023-11-27
Bump go version to 1.21

sayboras:
Relates: https://github.com/cilium/proxy/pull/439


go.mod
```diff
@@ -1,25 +1,25 @@
 module github.com/cilium/proxy
 
-go 1.20
+go 1.21
```
mhofstetter, line 3: we might want to set the version with the patch version

mhofstetter (approved):

go.mod
sayboras: ah good point :+1:

PR #447, opened 2023-12-02
github: Add CODEOWNERS file

sayboras:
This will help with review assignment automatically. We can start with the simple rule (e.g. cilium/proxy team is the owner), and improve if required later.

cc @cilium/proxy

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @cilium/envoy


pippolo84:
> 👍 IMO we should start with a dedicated group for envoy related stuff from the beginning (separated from FQDN)

I second Marco's suggestion. I think it'd be better to have two distinct groups.

mhofstetter (approved):
Thanks Tam!

PR #456, opened 2023-12-07
api: add proxy_id to bpf_metadata config and accesslog messages

jrajahalme:
[ upstream commit 94543384619f3de7f3110fb24fdc517822b721b4 ]

Add proxy_id to cilium metadata and accesslog messages. This is needed to correlate accesslog messages with a specific listener, when they operate on the same destination port number.


sayboras (approved):
LGTM :heavy_check_mark:

PR #502, opened 2024-01-22
api: add enable_reuse_port

mhofstetter:
Currently, the socket option `SO_REUSEPORT` is always set on the Listener sockets and is no longer configurable. Previously (before #315), this was configurable  via Envoy Listeners option `enable_reuse_port`.

Therefore, this commit introduces the config property `enable_reuse_port` on the `BPFMetadata` to reintroduce the possibility to configure this behaviour.


mhofstetter:
closing in favor of https://github.com/cilium/proxy/pull/505

PR #505, opened 2024-01-24
api: socket options shouldn't set SO_REUSEPORT on listener socket

mhofstetter:
Currently, the socket option `SO_REUSEPORT` is always set on all sockets - including the listener socket itself.

This doesn't take into account, that the Cilium Agent disables port reuse on the listener socket via Envoy listener configuration (`enable_reuse_port=false`) if BPF TPROXY is enabled.

In these cases, the cilium socket options
overrule and set the socket option `SO_REUSEPORT` - regardless of what's configured via Envoy listener API.. This leads to issues with BPF TPROXY (socket lookup fails).

Therefore, this commit changes the logic that the socket option `SO_REUSEPORT` is never set on the listener socket itself.

Note: The goal of #315 was to move privileged logic into the starter and remove the privileges from the Envoy process. But setting the socket option `SO_REUSEPORT` isn't a privileged operation. Thus it's ok to keep it configured via Envoy listener API.

Fixes: #315

### DIFF
--- a/cilium/socket_option.h
+++ b/cilium/socket_option.h
@@ -84,13 +84,15 @@ public:
         source_address = ipv4_source_address_;
       }
     }
+
+    uint32_t one = 1;
+
     // identity is zero for the listener socket itself, set transparent and reuse options also for
     // the listener socket.
     if (source_address || identity_ == 0) {
       // Allow reuse of the original source address. This may by needed for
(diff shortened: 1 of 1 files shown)


jrajahalme (approved):
Thanks for the fix!

PR #537, opened 2024-02-07
go: Bump golang to 1.22

sayboras:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/cilium/proxy
 
-go 1.21
+go 1.22
```
mhofstetter, line 3: why not pinning it to a specific patch version?

mhofstetter (approved):
:+1: thanks Tam

go.mod
sayboras: A few weeks back I was having issue with go version mismatch with cilium/cilium in CI, this is probably just a small fix, but I didn't look further. 

Having a generic 1.22 will work well with any go pinned version from cilium/cilium.

PR #564, opened 2024-02-19
patches: Remove duplicated upstream callbacks

sayboras:
This commit is to remove duplicated callback for method onUpstreamConnectionEstablished.

Relates: https://github.com/envoyproxy/envoy/blob/release/v1.28/source/common/router/upstream_request.cc#L682-L684
Relates: https://github.com/cilium/proxy/pull/324

Signed-off-by: Tam Mach <tam.mach@cilium.io>


jrajahalme (approved):
Thanks for the fix!

PR #568, opened 2024-02-21
docker: Add non-root user in Dockerfile

sayboras:
This is to improve the security posture of cilium/proxy running in the
deamonset mode.

Relates: https://github.com/cilium/cilium/issues/23217


sayboras:
We can put this one on hold due to the below 

https://github.com/kubernetes/website/pull/45150/commits/7fd32afce4d209775a2a9fe1e03abc3b20896368#r1504714307

PR #580, opened 2024-02-27
tests: Reduce flakiness

jrajahalme:
There is an inherent race between file-based policy update and the following test cases. Reduce flakiness by adding 100ms sleep after changing policy files.

### DIFF
--- a/tests/cilium_http_integration_test.cc
+++ b/tests/cilium_http_integration_test.cc
@@ -1201,6 +1201,9 @@ TEST_P(SDSIntegrationTest, TestMissingSDSSecretOnUpdate) {
                  "of using gRPC",
                  temp_path, policy_path);
 
+  // Reduce flakiness by allowing some time for the policy to be updated before the following test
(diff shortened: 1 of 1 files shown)


sayboras (approved):

tests/cilium_http_integration_test.cc
```diff
@@ -1201,6 +1201,9 @@ TEST_P(SDSIntegrationTest, TestMissingSDSSecretOnUpdate) {
                  "of using gRPC",
                  temp_path, policy_path);
 
+  // Reduce flakiness by allowing some time for the policy to be updated before the following test
+  absl::SleepFor(absl::Milliseconds(100));
```
sayboras, line 1205: is it easy to have this delay configured externally i.e. env, config, etc?

jrajahalme: Maybe, but hopefully that is not necessary, let's see how this goes first!

PR #649, opened 2024-03-18
cleanup: remove istio sidecar related code

mhofstetter:
We no longer build proxy images for istio, better just clean it up.

Hence, as an additional follow up to #448, this commit removes the Istio sidecar specific implementations (socket mark, ...).

### DIFF
--- a/cilium/socket_option.h
+++ b/cilium/socket_option.h
@@ -43,7 +43,7 @@ public:
 
   bool setOption(Network::Socket& socket,
                  envoy::config::core::v3::SocketOption::SocketState state) const override {
-    // sidecars do not have mark
+    // don't set option for mark 0 -> tests rely on this (they fail due to missing capabilities)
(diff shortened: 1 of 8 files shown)


cilium/socket_option.h
```diff
@@ -43,7 +43,7 @@ class SocketMarkOption : public Network::Socket::Option,
 
   bool setOption(Network::Socket& socket,
                  envoy::config::core::v3::SocketOption::SocketState state) const override {
-    // sidecars do not have mark
+    // don't set option for mark 0 -> tests rely on this (the fail due to missing capabilities)
```
mhofstetter, line 46: Without this, tests fail with

```
Failed to set socket option SO_MARK to 0, capability CAP_NET_ADMIN needed: Operation not permitted
```

jrajahalme (approved):
Thanks for the clean-up!

PR #659, opened 2024-03-20
bpfmetadata: refactor getMetadata by extracting methods

mhofstetter:
This PR refactors the function `BPFMetadata.getMetadata` by extracting logic into separate functions

- `resolveSourceIdentity`
- `getIPAddressPairFrom`
- `selectIPVersion`

In addition some small other refactorings were added. All of this should help to ease the understanding of this function and the BPFMetadata listener filter in general.

Please review the individual commits.

### DIFF
--- a/cilium/bpf_metadata.cc
+++ b/cilium/bpf_metadata.cc
@@ -238,19 +238,17 @@ Config::getIPAddressPairFrom(const Network::Address::InstanceConstSharedPtr sour
   return addressPair;
 }
 
-const Network::Address::Ip*
-Config::selectIPVersion(const Network::Address::IpVersion version,
-                        const Network::Address::InstanceConstSharedPtr ipv4SourceAddress,
-                        const Network::Address::InstanceConstSharedPtr ipv6SourceAddress) {
+const Network::Address::Ip* Config::selectIPVersion(const Network::Address::IpVersion version,
+                                                    const IPAddressPair& source_addresses) {
   switch (version) {
   case Network::Address::IpVersion::v4:
-    if (ipv4SourceAddress) {
-      return ipv4SourceAddress->ip();
+    if (source_addresses.ipv4_) {
+      return source_addresses.ipv4_->ip();
     }
     break;
   case Network::Address::IpVersion::v6:
-    if (ipv6SourceAddress) {
-      return ipv6SourceAddress->ip();
+    if (source_addresses.ipv6_) {
+      return source_addresses.ipv6_->ip();
     }
     break;
   }
@@ -314,8 +312,7 @@ bool Config::getMetadata(Network::ConnectionSocket& socket) {
   uint32_t ingress_source_identity = 0;
 
   // Use the configured IPv4/IPv6 Ingress IPs as starting point for the sources addresses
-  Network::Address::InstanceConstSharedPtr ipv4_source_address = ipv4_source_address_;
-  Network::Address::InstanceConstSharedPtr ipv6_source_address = ipv6_source_address_;
+  IPAddressPair source_addresses(ipv4_source_address_, ipv6_source_address_);
 
   // NOTE: As L7 LB does not use the original destination, there is a possibility of a 5-tuple
   // collision if the same source pod is communicating with the same backends on same destination
@@ -342,9 +339,7 @@ bool Config::getMetadata(Network::ConnectionSocket& socket) {
     // Keep the original source address for the matching IP version, create a new source IP for
     // the other version (with the same source port number) in case an upstream of a different
     // IP version is chosen.
-    IPAddressPair sourceIPs = getIPAddressPairFrom(src_address, policy->getEndpointIPs());
-    ipv4_source_address = sourceIPs.ipv4_;
-    ipv6_source_address = sourceIPs.ipv6_;
+    source_addresses = getIPAddressPairFrom(src_address, policy->getEndpointIPs());
 
     // Original source address is now in one of 'ipv[46]_source_address'
     src_address = nullptr;
@@ -353,8 +348,7 @@ bool Config::getMetadata(Network::ConnectionSocket& socket) {
     // if any and policy for this identity exists.
 
     // Pick the local source address of the same family as the incoming connection
-    const Network::Address::Ip* ip =
-        selectIPVersion(sip->version(), ipv4_source_address, ipv6_source_address);
+    const Network::Address::Ip* ip = selectIPVersion(sip->version(), source_addresses);
 
     if (!ip) {
       // IP family of the connection has no configured local source address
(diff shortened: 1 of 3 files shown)


cilium/bpf_metadata.h
```diff
@@ -48,6 +48,22 @@ class Config : public Cilium::PolicyResolver,
   Cilium::CtMapSharedPtr ct_maps_{};
   Cilium::IPCacheSharedPtr ipcache_{};
   std::shared_ptr<const Cilium::PolicyHostMap> hosts_{};
+
+private:
+  uint32_t resolveSourceIdentity(const PolicyInstanceConstSharedPtr policy,
+                                 const Network::Address::Ip* sip, const Network::Address::Ip* dip,
+                                 bool ingress, bool isL7LB);
+
+  void getL7LBEastWestSourceIPs(const PolicyInstanceConstSharedPtr policy,
```
mhofstetter, line 57: TODO: use new struct to return the IPs instead of passing them as parameters

cilium/bpf_metadata.cc
```diff
@@ -208,6 +208,64 @@ uint32_t Config::resolveSourceIdentity(const PolicyInstanceConstSharedPtr policy
...
+  case Network::Address::IpVersion::v6:
+    if (ipv6SourceAddress) {
+      ip = ipv6SourceAddress->ip();
+    }
+    break;
+  }
+
+  return ip;
+}
+
```
jrajahalme, line 268: This could be more generic `selectIPVersion()`, if the first argument would be `Network::Address::IpVersion`, i.e.:
```
const Network::Address::Ip*
Config::selectIPVersion(Network::Address::IpVersion version,
                        Network::Address::InstanceConstSharedPtr ipv4Address,
                        Network::Address::InstanceConstSharedPtr ipv6Address) {
  ...
```

cilium/bpf_metadata.cc
```diff
@@ -208,6 +208,64 @@ uint32_t Config::resolveSourceIdentity(const PolicyInstanceConstSharedPtr policy
...
+    if (ips.ipv4_) {
+      sockaddr_in sa4 = *reinterpret_cast<const sockaddr_in*>(ips.ipv4_->sockAddr());
+      sa4.sin_port = htons(sip->port());
+      sourceIPs.ipv4 = std::make_shared<Network::Address::Ipv4Instance>(&sa4);
+    }
+    break;
+  }
+
+  return sourceIPs;
+}
```
jrajahalme, line 242: In general, I'd like to see factored-out functions as "generic utilities", while the business logic would be kept on the caller. Here "east/west" vs. "north/south" parlance would be part of the business logic.

To that end, consider:
- reuse `IPAddressPair` type, and take reference to it as a parameter instead of `policy` (so the `policy->getEndpointIPs()` call would be on the caller side)
- this function keeps the source address, and fills in the other address version from the given `IPAddressPair`, so maybe call this `IPAddressPair getIPAddressPairFrom(const Network::Address::Ip* sip, const IPAddressPair& addresses)`?
- add a comment above explaining the above

cilium/bpf_metadata.cc
```diff
@@ -295,52 +353,16 @@ bool Config::getMetadata(Network::ConnectionSocket& socket) {
...
-          ipv4_source_address = std::make_shared<Network::Address::Ipv4Instance>(&sa4);
-        } else {
-          ipv4_source_address = nullptr;
-        }
-      } break;
-      }
+      SourceIPs sourceIPs = getL7LBEastWestSourceIPs(policy, sip, src_address);
+      ipv4_source_address = sourceIPs.ipv4;
+      ipv6_source_address = sourceIPs.ipv6;
+
```
jrajahalme, line 353: This would be simpler if we kept the addresses in an `IPAddressPair` type instead, then this could be just
```
sourceAddresses = getIPAddressPairFrom(sip, policy->getEndpointIPs());
```

jrajahalme (commented):
Some suggestions to make the factored out functions be more "generic utilities" while keeping the business logic on the caller side.

cilium/bpf_metadata.cc
mhofstetter: good point. i changed it accordingly.

cilium/bpf_metadata.cc
mhofstetter: thanks. changed

cilium/bpf_metadata.cc
mhofstetter: added an additional commit that replaces the two variables with one of type `IPAddressPair`. This way we can assign it directly.

mhofstetter:
> Some suggestions to make the factored out functions be more "generic utilities" while keeping the business logic on the caller side.

Thanks Jarno - I incorporated your feedback! PTAL.

jrajahalme (changes requested):
Some suggestions for the last commit, tested locally with these changes:
```
diff --git a/cilium/bpf_metadata.cc b/cilium/bpf_metadata.cc
index 0d6a09993..7cec1450c 100644
--- a/cilium/bpf_metadata.cc
+++ b/cilium/bpf_metadata.cc
@@ -239,7 +239,7 @@ Config::getIPAddressPairFrom(const Network::Address::InstanceConstSharedPtr sour
 }
 
 const Network::Address::Ip* Config::selectIPVersion(const Network::Address::IpVersion version,
-                                                    const IPAddressPair sourceAddresses) {
+                                                    const IPAddressPair& sourceAddresses) {
   switch (version) {
   case Network::Address::IpVersion::v4:
     if (sourceAddresses.ipv4_) {
@@ -312,9 +312,7 @@ bool Config::getMetadata(Network::ConnectionSocket& socket) {
   uint32_t ingress_source_identity = 0;
 
   // Use the configured IPv4/IPv6 Ingress IPs as starting point for the sources addresses
-  IPAddressPair sourceAddresses = IPAddressPair();
-  sourceAddresses.ipv4_ = ipv4_source_address_;
-  sourceAddresses.ipv6_ = ipv6_source_address_;
+  IPAddressPair sourceAddresses(ipv4_source_address_, ipv6_source_address_);
 
   // NOTE: As L7 LB does not use the original destination, there is a possibility of a 5-tuple
   // collision if the same source pod is communicating with the same backends on same destination
diff --git a/cilium/bpf_metadata.h b/cilium/bpf_metadata.h
index 71c5c5b46..4139b7834 100644
--- a/cilium/bpf_metadata.h
+++ b/cilium/bpf_metadata.h
@@ -58,7 +58,7 @@ private:
                                      const IPAddressPair& addresses);
 
   const Network::Address::Ip* selectIPVersion(const Network::Address::IpVersion version,
-                                              const IPAddressPair sourceAddresses);
+                                              const IPAddressPair& sourceAddresses);
 };
 
 typedef std::shared_ptr<Config> ConfigSharedPtr;
diff --git a/cilium/network_policy.h b/cilium/network_policy.h
index 87857038d..8bf3395f3 100644
--- a/cilium/network_policy.h
+++ b/cilium/network_policy.h
@@ -90,6 +90,9 @@ private:
 class IPAddressPair {
 public:
   IPAddressPair(){};
+  IPAddressPair(Network::Address::InstanceConstSharedPtr& ipv4,
+                Network::Address::InstanceConstSharedPtr& ipv6)
+      : ipv4_(ipv4), ipv6_(ipv6){};
   IPAddressPair(const cilium::NetworkPolicy& proto);
 
   Network::Address::InstanceConstSharedPtr ipv4_{};
```

mhofstetter:
> Some suggestions for the last commit, tested locally with these changes:

Thanks @jrajahalme - i applied your patch.

jrajahalme (approved):

Issue #470, opened 2024-01-08
Backend Connectivity Issues


TECHNOFAB11:
I've been having issues with the Envoy Gateway and Ingress with cilium for some months now. I'm not sure if it's cilium's or envoy's fault, envoy logs "conntrack lookup failed: Success" on every request (it always returns an error because it does not receive a response from the backend) but it's just an info log.

My issue in cilium/cilium with infos etc.:
https://github.com/cilium/cilium/issues/29406

TECHNOFAB11:
I tested around and installed Envoy Gateway and some more GatewayAPI / Ingress controllers and they all worked, so this is probably not Envoy's fault. 
The "conntrack lookup failed: Success" is weird though, so I'll keep this open :D

PR #674, opened 2024-03-26
bpf: fix error propagation for bpf syscalls

mhofstetter:
Currently, interaction with BPF maps via syscalls (open, lookup) might result in log messages of the following form, where the error detail is `success`:

```
[info][filter] [cilium/conntrack.cc:229] cilium.bpf_metadata: IPv4 conntrack map global lookup failed: Success
```

This is due to the fact that BPF maps are accessed in the starter process. Hence, the syscalls are also executed in this separate process and the variable `errno` is never set in the Envoy process where the log is written..

Therefore, this commit fixes the error propagation by setting the variable `errno` after retrieving the response from the privileged client doing the call to the starter process.

Fixes: #315
Related to: #470

### DIFF
--- a/cilium/bpf.cc
+++ b/cilium/bpf.cc
@@ -1,5 +1,7 @@
 #include "cilium/bpf.h"
 
+#include <errno.h>
+
 #include "source/common/common/utility.h"
 
 #include "cilium/privileged_service_client.h"
@@ -92,12 +94,21 @@ bool Bpf::open(const std::string& path) {
               Envoy::errorDetails(ret.errno_));
   }
 
+  errno = ret.errno_;
+
   return false;
 }
 
 bool Bpf::lookup(const void* key, void* value) {
   auto& cilium_calls = PrivilegedService::Singleton::get();
-  return cilium_calls.bpf_lookup(fd_, key, key_size_, value, value_size_).return_value_ == 0;
+  auto result = cilium_calls.bpf_lookup(fd_, key, key_size_, value, value_size_);
+
+  if (result.return_value_ == 0) {
+    return true;
+  }
+
+  errno = result.errno_;
+  return false;
 }
 
 } // namespace Cilium


mhofstetter:
> If this fixes https://github.com/cilium/proxy/issues/470, does it also fix https://github.com/cilium/cilium/issues/29406 and https://github.com/cilium/cilium/issues/30510 ? I believe some of those are related (or I'm hitting more of them in my setup)

https://github.com/mhofstetter/proxy/commit/c73651f124cf7c4226f4ed9820c0df50d4a5fc85#commitcomment-140253867

@zviratko I'm answering here in the PR instead of in the comment in the commit for more visibility.

No, sorry - this PR only fixes the error propagation that currently hides the underlying error from being shown in the log message and lead to the somewhat special error message mentioned in the [second comment of the issue](https://github.com/cilium/proxy/issues/470#issuecomment-1900216552).

> The "conntrack lookup failed: Success" is weird though, so I'll keep this open :D

cilium/bpf.cc
```diff
@@ -87,17 +89,24 @@ bool Bpf::open(const std::string& path) {
   } else if (ret.errno_ == ENOENT && log_on_error) {
     ENVOY_LOG(debug, "cilium.bpf_metadata: bpf syscall for map {} failed: {}", path,
               Envoy::errorDetails(ret.errno_));
+    errno = ret.errno_;
   } else if (log_on_error) {
     ENVOY_LOG(warn, "cilium.bpf_metadata: bpf syscall for map {} failed: {}", path,
               Envoy::errorDetails(ret.errno_));
+    errno = ret.errno_;
   }
```
jrajahalme, line 98: It seems to me that since the if blocks are conditioned on `log_on_error`, we are missing the error cases where `log_on_error == false`. Maybe simply set the errno here, just before `return false;`?

jrajahalme (changes requested):

cilium/bpf.cc
mhofstetter: ouch right. fixed. thanks

PR #708, opened 2024-04-15
api: Remove explicit go get in favor of Renovate

jrajahalme:
We have Renovate updating Go dependencies now, so we do not need to track them manually in Makefile.api any more. After this change `make api` does not downgrade the dependencies any more.


sayboras (approved):
this is overlook for a long time, probably we need to run `make api` again.

PR #711, opened 2024-04-18
make: Install clang without checking /usr/lib/llvm-17

jrajahalme:
/usr/lib/llvm-17 may exist while all the clang dependencies may not be installed. Remove the '/usr/lib/llvm-17' target, and install clang on the 'clang.bazelrc' target directly. After this clang is installed whenever clang.bazelrc is missing, or bazel/setup_clang.sh changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,29 +62,6 @@ else
   BAZEL_BUILD_OPTS += --config=release
 endif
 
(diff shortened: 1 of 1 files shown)


sayboras (approved):
Still good ✅

PR #727, opened 2024-04-24
[v1.28] Backport of #650

mhofstetter:
### starter: introduce possibility to keep capability NET_BIND_SERVICE

Currently, the "starter" drops all capabilities before starting the Envoy process itself. This is mainly to prevent the Envoy process from having the capabilities NET_ADMIN, SYS_ADMIN etc.

But this change also comes with the drawback that it prevents Envoy from binding to privileged ports - because the capability NET_BIND_SERVICE gets dropped as well.

Therefore, this commit introduces a new flag --keep-cap-net-bind-service to the starter. If this flag is present, the capability NET_BIND_SERVICE is kept in the privileged and effective set for the Envoy process.

Backport of https://github.com/cilium/proxy/pull/650

### DIFF
--- a/cilium/privileged_service_client.cc
+++ b/cilium/privileged_service_client.cc
@@ -10,7 +10,10 @@ namespace PrivilegedService {
 
 ProtocolClient::ProtocolClient()
     : Protocol(CILIUM_PRIVILEGED_SERVICE_FD), call_mutex_(PTHREAD_MUTEX_INITIALIZER), seq_(0) {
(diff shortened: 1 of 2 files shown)


mhofstetter:
@sayboras 

is the compatibility matrix still up-to-date? `main (cilium/cilium) -> v1.28.x`?

or is this backport not necessary at all because we target 1.29?

sayboras:
Yes, backport is not required right now. Thanks for pointing it out, I will update the matrix :+1:

PR #728, opened 2024-04-24
docs: Update version matrix

sayboras:
The main point is that upcoming v1.15.5 will be the first release version with envoy v1.28.x, and main cilium/cilium branch is now with v1.29.x.

Suggested-by: Marco Hofstetter <marco.hofstetter@isovalent.com>


mhofstetter (approved):
Thanks Tam!

PR #742, opened 2024-05-02
bpf_metadata: Use original source address even if the destination is external

jrajahalme:
Use the original source address if permitted even if the destination is external to the cluster.

### DIFF
--- a/cilium/bpf_metadata.cc
+++ b/cilium/bpf_metadata.cc
@@ -384,12 +384,9 @@ bool Config::getMetadata(Network::ConnectionSocket& socket) {
     // This means that a local host IP is used if no IP is configured to be used instead of it
     // ('ip' above is null).
     src_address = nullptr;
-  } else if (!(use_original_source_address_ &&
-               !(destination_identity & Cilium::ID::LocalIdentityFlag) &&
-               destination_identity != Cilium::ID::WORLD && !npmap_->exists(other_ip))) {
-    // Otherwise only use the original source address if permitted, destination identity is not a
-    // locally allocated identity, is not classified as WORLD, and the destination is not in the
-    // same node.
+  } else if (!use_original_source_address_ || npmap_->exists(other_ip)) {
+    // Otherwise only use the original source address if permitted and the destination is not
+    // in the same node.
 
     // Original source address is not used
     src_address = nullptr;


cilium/bpf_metadata.cc
```diff
@@ -384,12 +384,9 @@ bool Config::getMetadata(Network::ConnectionSocket& socket) {
     // This means that a local host IP is used if no IP is configured to be used instead of it
     // ('ip' above is null).
     src_address = nullptr;
-  } else if (!(use_original_source_address_ &&
-               !(destination_identity & Cilium::ID::LocalIdentityFlag) &&
-               destination_identity != Cilium::ID::WORLD && !npmap_->exists(other_ip))) {
-    // Otherwise only use the original source address if permitted, destination identity is not a
-    // locally allocated identity, is not classified as WORLD, and the destination is not in the
-    // same node.
+  } else if (!use_original_source_address_ || npmap_->exists(other_ip))) {
```
mhofstetter, line 387: `||` -> `&&` ?

cilium/bpf_metadata.cc
```diff
@@ -384,12 +384,9 @@ bool Config::getMetadata(Network::ConnectionSocket& socket) {
...
+    // Otherwise only use the original source address if permitted and the destination is not
+    //in the same node.
```
mhofstetter, line 389: nit: `//in` -> `// in` (add space after slash)

cilium/bpf_metadata.cc
jrajahalme: I inverted the logic via de Morgan's laws to simplify the logic (outer `not` is applied to each element inside the brackets and each `and`/`or` is flipped to `or`/`and`).

mhofstetter: ah right, i overlooked that it's not only the `use_original_source_address` that was inverted. I kind of gave the reason for this simplification just by asking and being able to read it correctly :smile:

cilium/bpf_metadata.cc
jrajahalme: fixed

cilium/bpf_metadata.cc
```diff
@@ -384,12 +384,9 @@ bool Config::getMetadata(Network::ConnectionSocket& socket) {
...
     // ('ip' above is null).
     src_address = nullptr;
-  } else if (!(use_original_source_address_ &&
-               !(destination_identity & Cilium::ID::LocalIdentityFlag) &&
-               destination_identity != Cilium::ID::WORLD && !npmap_->exists(other_ip))) {
-    // Otherwise only use the original source address if permitted, destination identity is not a
-    // locally allocated identity, is not classified as WORLD, and the destination is not in the
-    // same node.
+  } else if (!use_original_source_address_ || npmap_->exists(other_ip)) {
+    // Otherwise only use the original source address if permitted and the destination is not
```
mhofstetter, line 388: nit: the comment reads a little bit strange. `Otherwise` would more relate to the `else` case (instead of `else if`).

I would rephrase this to something like this (to help documenting the actual condition)

> Don't use the original source address if not configured or the destination is on the same node.

jschwinger233:
Hi @jrajahalme , may I confirm as of this PR, proxy uses original src IP even for go-to-world traffic, like pod to 1.1.1.1 ?

PR #778, opened 2024-05-22
envoy: Patch original destination cluster to handle parallel host instantiation

jrajahalme:
Original destination cluster creates Host instances for the destination addresses on-demand. When multiple worker threads need to forward to the same destination at the same time (as is possible with `curl --parallel`, for example), separate instances of the destination Host objects are created. Meanwhile, Envoy uses `HostSharedPtr` as a map key in connection pool containers, which leads to multiple different connections being created for the same destination, even from the same worker, as the originally used Host instance is replaced with another, depending on which worker thread gets to add their Host object to the shared cluster state first. Multiple connections to the same destination fail when the original source address and port is being used, as is typically needed for Cilium policy enforcement, due to socket bind failing with `cannot bind requested address` error.

This PR fixes this by modifying the original destination cluster implementation to only create one Host object for any destination for any one cluster instance. This requires closer coordination between the worker threads in a form of an additional mutex, on which a write lock is taken whenever a new Host needs to be created. This synchronizes creation of new Hosts between the threads so that a new Host instance is created only if one has not yet been created by any of the worker threads.

Another possible way of fixing this issue would be to replace `HostSharedPtr` with the destination address as a map key, wherever it is used. This strategy could fix related issues with other cluster and/or load balancer types, it they were to exist. Upstream discussion is needed to settle on the proper fix, and we can adapt as needed when that is done.

It is yet to be decided if and when this should be backported.

### DIFF
--- a/patches/0006-original_dst_cluster-Avoid-multiple-hosts-for-the-sa.patch
+++ b/patches/0006-original_dst_cluster-Avoid-multiple-hosts-for-the-sa.patch
@@ -1,0 +1,458 @@
+From 6bbf94884ee4ec2b434f5d265b154d778df53d77 Mon Sep 17 00:00:00 2001
+From: Jarno Rajahalme <jarno@isovalent.com>
+Date: Fri, 24 May 2024 18:27:28 +0200
+Subject: [PATCH 6/6] original_dst_cluster: Avoid multiple hosts for the same
+ address
+
+Connection pool containers use HostSharedPtr as map keys, rather than the
+address of the host. This leads to multiple connections when there are
+multiple Host instances for the same address. This is breaking use of the
+original source address and port for upstream connections since only one
+such connection can exist at any one time.
+
+Original destination cluster implementation creates such duplicate Host
+instances when two worker threads are racing to create a Host for the
+same destination at the same time.
+
+Fix this by keeping a separate 'updates_map' where each worker places a
+newly created Host for the original destination. This map is used to look
+for the Host is it can not be found from the shared read-only
+'host_map'. Access to 'updates_map' is syncronized so that it can be
+safely shared by the worker threads. The main threads consolidates the
+updates from the 'updates_map' to a new instance of the shared, read-only
+hosts map, so that the workers do not need to stall for possibly large
+map updates.
+
+Signed-off-by: Jarno Rajahalme <jarno@isovalent.com>
+
+diff --git a/source/extensions/clusters/original_dst/original_dst_cluster.cc b/source/extensions/clusters/original_dst/original_dst_cluster.cc
+index 2b33138629..723b236d83 100644
+--- a/source/extensions/clusters/original_dst/original_dst_cluster.cc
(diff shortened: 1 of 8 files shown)


patches/0006-original_dst_cluster-Avoid-multiple-hosts-for-the-sa.patch
```diff
@@ -0,0 +1,465 @@
...
+ };
+ 
+-using HostsForAddressSharedPtr = std::shared_ptr<HostsForAddress>;
+-using HostMultiMap = absl::flat_hash_map<std::string, HostsForAddressSharedPtr>;
+-using HostMultiMapSharedPtr = std::shared_ptr<HostMultiMap>;
+-using HostMultiMapConstSharedPtr = std::shared_ptr<const HostMultiMap>;
++using HostUseSharedPtr = std::shared_ptr<HostUse>;
++using HostUseMap = absl::flat_hash_map<std::string, HostUseSharedPtr>;
++using HostUseMapUniquePtr = std::unique_ptr<HostUseMap>;
++using HostUseMapSharedPtr = std::shared_ptr<HostUseMap>;
```
sayboras, line 401: this is un-used

patches/0006-original_dst_cluster-Avoid-multiple-hosts-for-the-sa.patch
```diff
@@ -0,0 +1,465 @@
...
+       // 3) will not delete h since it takes at least one cleanup_interval for
+       // the host to set used_ bit for h to false.
+       bool keep = false;
+-      if (hosts->used_) {
+-        keep = true;
+-        hosts->used_ = false; // Mark to be removed during the next round.
++      if (host_use->used_) {
++	keep = true;
++	host_use->used_ = false; // Mark to be removed during the next round.
+       } else if (Runtime::runtimeFeatureEnabled(
```
sayboras, line 289: should we enable this runtime flag in cilium repo?

sayboras (approved):
Do we need to enable related upstream tests to make sure that we didn't break anything ?

patches/0006-original_dst_cluster-Avoid-multiple-hosts-for-the-sa.patch
jrajahalme: I was confused about this also, but runtime flags defined like this one are `true` by default.

patches/0006-original_dst_cluster-Avoid-multiple-hosts-for-the-sa.patch
jrajahalme: removed, thanks!

patches/0006-original_dst_cluster-Avoid-multiple-hosts-for-the-sa.patch
jrajahalme: https://github.com/envoyproxy/envoy/blob/main/source/common/runtime/runtime_features.cc#L25-L27

PR #785, opened 2024-05-30
Ipcache reopen on cilium restart

jrajahalme:
Reopen ipcache whenever a new NetworkPolicy stream is opened. This way we get access to the new ipcache after cilium agent restart.

### DIFF
--- a/cilium/network_policy.cc
+++ b/cilium/network_policy.cc
@@ -1063,20 +1063,17 @@ private:
 // Common base constructor
 // This is used directly for testing with a file-based subscription
 NetworkPolicyMap::NetworkPolicyMap(Server::Configuration::FactoryContext& context)
-    : tls_map_(context.serverFactoryContext().threadLocal()),
-      local_ip_str_(context.serverFactoryContext().localInfo().address()->ip()->addressAsString()),
+    : context_(context.serverFactoryContext()), tls_map_(context_.threadLocal()),
+      local_ip_str_(context_.localInfo().address()->ip()->addressAsString()),
       name_(fmt::format("cilium.policymap.{}.{}.", local_ip_str_, ++instance_id_)),
-      scope_(context.serverFactoryContext().serverScope().createScope(name_)),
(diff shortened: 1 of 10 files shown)


mhofstetter (approved):
LGTM - thanks for the detailed commit messages!

Does the same issue exist with the conntrack map? (or does the Agent behaves differently for this map(s) (global & local)?

cilium/network_policy.cc
```diff
@@ -1274,10 +1284,22 @@ NetworkPolicyMap::onConfigUpdate(const std::vector<Envoy::Config::DecodedResourc
...
 
+    // Reopen IPcache for every new stream. Cilium agent re-creates IP cache on restart,
+    // and that is also when the old stream terminates and a new one is created.
+    // New security identities (e.g., for FQDN policies) only get inserted to the new IP cache,
+    // so open it before the workers get a chance to enforce policy on the new IDs.
+    if (isNewStream()) {
+      IPCacheSharedPtr ipcache = IPCache::GetIPCache(context_);
+      if (ipcache != nullptr) {
+        ENVOY_LOG(trace, "Reopening ipcache on new stream");
+        ipcache->Open();
```
sayboras, line 1299: should we do the same check and reset if there is any issue with opening file, also in that case, might need to start subscription for hostmap as well.

jrajahalme: I think hostmap gets little testing coverage nowadays, would not want to change the operating mode now, on the contrary, I'd like to retire the hostmap ASAP.

But we need to handle the failure case here somehow. I'll try making the lookup reopen when needed.

jrajahalme: Updated the 1st commit to keep the path and retry open at the beginning of lookup, if it failed. This way we'll recover if there is some temporary failure on opening the map.

jrajahalme:
Also had to move the reopen to happen even if there is no difference in the policy!

sayboras (approved):

Issue #829, opened 2024-07-03
Bump to envoy version to 1.30


sayboras:
Step by step guide can be found in the below docs

https://github.com/cilium/proxy/blob/main/UPGRADE_ENVOY.md

PR #831, opened 2024-07-03
envoy: Bump envoy version to v1.30.x

sayboras:
As v1.16 branch is cut in cilium/cilium, we can start upgrading envoy to v1.30.x for main. Testing was done in upstream https://github.com/cilium/cilium/pull/33607

Fixes: https://github.com/cilium/proxy/issues/829


patches/0001-network-Add-callback-for-upstream-authorization.patch
```diff
@@ -289,7 +289,7 @@ index 9bc074f393..8d035b5cad 100644
    if (generic_conn_pool_) {
 +    bool accepted = read_callbacks_->iterateUpstreamCallbacks(generic_conn_pool_->host(), getStreamInfo());
 +    if (!accepted) {
-+      getStreamInfo().setResponseFlag(StreamInfo::ResponseFlag::UnauthorizedExternalService);
++      getStreamInfo().setResponseFlag(StreamInfo::CoreResponseFlag::UnauthorizedExternalService);
```
jrajahalme, line 292: This change could be folded in to the earlier commit updating the patches?

jrajahalme (approved):
Great work! Only one minor nit about patch update in another commit!

patches/0001-network-Add-callback-for-upstream-authorization.patch
sayboras: good catch, thanks.

PR #850, opened 2024-07-21
bazel: Switch to bazelisk

sayboras:
This is to avoid manually detecting bazel version, and also to improve user experience while working on multiple release branches.

Relates: https://github.com/bazelbuild/bazelisk


tools/install_bazel.sh
```diff
@@ -21,6 +20,6 @@ else
     OS=$(uname -s | tr '[:upper:]' '[:lower:]')
     ARCH=$(uname -m) && [ "$ARCH" != "aarch64" ] || ARCH="arm64"
     echo "Downloading bazel-${BAZEL_VERSION}-${OS}-${ARCH} to ${BAZEL}"
```
jrajahalme, line 22: This should be updated to match the actual download.

Makefile
```diff
@@ -92,9 +92,9 @@ else
   include Makefile.docker
 
   # Fetch and install Bazel if needed
-  .PHONY: install-bazel
-  install-bazel:
-	tools/install_bazel.sh `cat .bazelversion`
+  .PHONY: install-bazelisk
+  install-bazelisk:
+	tools/bazelisk.sh
```
jrajahalme, line 97: Is `tools/baselisk.sh` already in the code base?

tools/install_bazel.sh
```diff
@@ -1,13 +1,12 @@
...
 
-BAZEL_VERSION=$1
 if ! command -v sudo >/dev/null; then
     SUDO=
 else
     SUDO=sudo
 fi
 
-echo "Checking if Bazel ${BAZEL_VERSION} needs to be installed..."
+echo "Checking if Bazelisk ${BAZEL_VERSION} needs to be installed..."
```
jrajahalme, line 9: Does it make sense to echo `${BAZEL_VERSION}` if it is not defined?

tools/install_bazel.sh
```diff
@@ -21,6 +20,6 @@ else
     OS=$(uname -s | tr '[:upper:]' '[:lower:]')
     ARCH=$(uname -m) && [ "$ARCH" != "aarch64" ] || ARCH="arm64"
     echo "Downloading bazel-${BAZEL_VERSION}-${OS}-${ARCH} to ${BAZEL}"
-    ${SUDO} curl -sfL "https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-${OS}-${ARCH}" -o ${BAZEL}
-    ${SUDO} chmod +x ${BAZEL}
+    ${SUDO} curl -sfL "https://github.com/bazelbuild/bazelisk/releases/download/${BAZEL_VERSION}/bazel-${OS}-${ARCH}" -o "${BAZEL}"
+    ${SUDO} chmod +x "${BAZEL}"
```
jrajahalme, line 24: Are we actually running this (vs. `tools/bazelisk.sh`?)

jrajahalme (changes requested):
Seems we are missing `tools/bazelisk.sh` and/or not using `tools/install_bazel.sh` any more?

Makefile
sayboras: good catch, thanks.

tools/install_bazel.sh
sayboras: thanks, I have added BAZELISK_VERSION along with renovate auto upgrade.

tools/install_bazel.sh
sayboras: renamed the script :heavy_check_mark:

tools/install_bazel.sh
sayboras: I have updated the script to lock down the bazelisk version, bazel version will be just auto detected by bazelisk.

tools/install_bazelisk.sh
```diff
@@ -0,0 +1,28 @@
...
+else
+    BAZEL=$(command -v bazel)
+    if [ -n "${BAZEL}" ] ; then
+        echo "Removing old Bazel version at ${BAZEL}"
+        ${SUDO} rm "${BAZEL}"
+    else
+        BAZEL=/usr/local/bin/bazel
+    fi
+    OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+    ARCH=$(uname -m) && [ "$ARCH" != "x86_64" ] || ARCH="amd64"
```
jrajahalme, line 24: This will not work for `uname -m` returning `aarch64`, which will need to be translated to `arm64`, like we did for `install_bazel.sh`. Seems `bazelisk` needs `x86_64` to be translated to `amd64` (like you did here) while `bazel` did not.

jrajahalme (changes requested):

tools/install_bazelisk.sh
sayboras: Updated and thanks.

tools/install_bazelisk.sh
```diff
@@ -0,0 +1,33 @@
...
```
jrajahalme, line 24: Now that `ARCH` versions are handled below this should be just:
```
ARCH=$(uname -m)
```

jrajahalme (changes requested):

jrajahalme (approved):
Thanks, this makes life easier going forward!

PR #890, opened 2024-08-14
tls_wrapper: Do not use raw socket if no policy is found

jrajahalme:
Currently TLS wrapper is only used when the policy should also specify a TLS context for the connection. Therefore we should never use a raw socket instead when the TLS context is not available.


sayboras (approved):
LGTM, might need to backport this to older branches as well.

PR #917, opened 2024-08-31
policy: Use timing-safe string comparison

sayboras:
When validating the input header value against a secret, it is advisable to use a timing-safe string comparison operator with a runtime that is independent of the number of equivalent prefix bytes of the secret and header value.  This commit is to use double-hmac method to compare secret values.


sayboras:
Closed in favor of https://github.com/cilium/proxy/pull/974

Issue #958, opened 2024-10-03
NetworkPolicy xDS stream stalls for ~20 seconds before finishing updates on all worker threads


jrajahalme:
NPDS version 247 is received at 07:14:02, some worker threads update immediately after, but other only at 07:14:22, causing Cilium Agent to not be able to get ACKs on policy updates in time (100ms timeout is much less than 20 seconds):

from `logs-cilium-envoy-kgzx9-cilium-envoy-20241003-071420.log`:
```
2024-10-03T07:14:02.719903223Z [2024-10-03 07:14:02.719][11][debug][config] [external/envoy/source/extensions/config_subscription/grpc/grpc_mux_impl.cc:268] Received gRPC message for type.googleapis.com/cilium.NetworkPolicy at version 247
2024-10-03T07:14:02.720433651Z [2024-10-03 07:14:02.719][11][debug][config] [cilium/network_policy.cc:1200] Received Network Policy for endpoint 122 in onConfigUpdate() version 247
2024-10-03T07:14:02.720443269Z [2024-10-03 07:14:02.719][11][debug][config] [cilium/network_policy.cc:1200] Received Network Policy for endpoint 2491 in onConfigUpdate() version 247
2024-10-03T07:14:02.720451755Z [2024-10-03 07:14:02.719][11][debug][config] [cilium/network_policy.cc:1200] Received Network Policy for endpoint 1166 in onConfigUpdate() version 247
2024-10-03T07:14:02.720461132Z [2024-10-03 07:14:02.719][11][debug][config] [cilium/network_policy.cc:1200] Received Network Policy for endpoint 2283 in onConfigUpdate() version 247
2024-10-03T07:14:02.720470640Z [2024-10-03 07:14:02.719][11][debug][config] [cilium/network_policy.cc:1200] Received Network Policy for endpoint 3851 in onConfigUpdate() version 247
2024-10-03T07:14:02.720545489Z [2024-10-03 07:14:02.719][11][debug][config] [cilium/network_policy.cc:1200] Received Network Policy for endpoint 3704 in onConfigUpdate() version 247
2024-10-03T07:14:02.720578060Z [2024-10-03 07:14:02.719][11][debug][config] [cilium/network_policy.cc:1200] Received Network Policy for endpoint 4060 in onConfigUpdate() version 247
[...]

PR #959, opened 2024-10-03
policy: Add warning log if worker threads take longer than 100ms to update policy maps

jrajahalme:
Add a warning log if worker threads take longer than 100ms to update their network policy maps. This is to help identify cases like #958, where worker thread [20] took almost 20 seconds to start running the update.

Related: #958


sayboras (approved):

jrajahalme:
Fixed format.

PR #974, opened 2024-10-17
policy: Use timing-safe string comparison

sayboras:
When validating the input header value against a secret, it is advisable to use a timing-safe string comparison operator with a runtime independent of the number of equivalent prefix bytes of the secret and header value. This commit is to use CRYPTO_memcmp function[^1][^2] for string comparison, similar to what is done in envoy codebase.

[^1]: https://github.com/openssl/openssl/blob/master/doc/man3/CRYPTO_memcmp.pod
[^2]: https://github.com/openssl/openssl/blob/f4c467452694e1211395d17c2c027d99c35ee1e1/crypto/cpuid.c#L184

### DIFF
--- a/cilium/network_policy.cc
+++ b/cilium/network_policy.cc
@@ -18,6 +18,7 @@
 #include "cilium/grpc_subscription.h"
 #include "cilium/ipcache.h"
 #include "cilium/secret_watcher.h"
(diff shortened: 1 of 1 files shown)


jrajahalme (approved):
Looks good, thanks for digging in :-)

PR #987, opened 2024-10-26
envoy: Enable all extensions

sayboras:
Just wanna check the build time and also binary size, the current binary is around ~55MB.

New build is taking almost same time (~10 mins), probably due to caching, but the binary size is increasing to 68M

```
root@75e5b8c73034:/# ls -lrth /usr/bin/cilium-envoy
-rwxr-xr-x 1 1337 1337 68M Oct 26 15:03 /usr/bin/cilium-envoy
```

### DIFF
--- a/envoy_build_config/extensions_build_config.bzl
+++ b/envoy_build_config/extensions_build_config.bzl
@@ -17,10 +17,10 @@ EXTENSIONS = {
     # Clusters
     #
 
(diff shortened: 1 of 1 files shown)


jrajahalme:
Might as well do this then? I guess the biggest build time penalty was from WASM that is already enabled?

PR #995, opened 2024-10-31
policy: Add stats

jrajahalme:
Add policy stats for number of updates, rejected updates, and update time limit timeouts, as well as a histogram for the update time duration.

Make the update time limit configurable via BfpMetadata filter config, as the policy map is created from there. Default is the same as in Cilium Agent (100ms).

### DIFF
--- a/cilium/l7policy.h
+++ b/cilium/l7policy.h
@@ -18,8 +18,8 @@ namespace Cilium {
  * All Cilium L7 filter stats. @see stats_macros.h
  */
 // clang-format off
-#define ALL_CILIUM_STATS(COUNTER)                                                                  \
-  COUNTER(access_denied)                                                                                \
+#define ALL_CILIUM_STATS(COUNTER)	\
+  COUNTER(access_denied)
 // clang-format on
 
 /**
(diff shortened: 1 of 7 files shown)


cilium/network_policy.cc
```diff
@@ -1080,6 +1081,9 @@ NetworkPolicyMap::NetworkPolicyMap(Server::Configuration::FactoryContext& contex
       local_ip_str_(context_.localInfo().address()->ip()->addressAsString()),
       name_(fmt::format("cilium.policymap.{}.{}.", local_ip_str_, ++instance_id_)),
       scope_(context_.serverScope().createScope(name_)),
+      stats_{ALL_CILIUM_POLICY_STATS(POOL_COUNTER_PREFIX(*scope_, "policy."),
```
mhofstetter, line 1084: can you list an example of how this looks like in the prometheus output?

i wonder whether it would make sense to have configurable stats_prefix like it's often the case for envoy metrics.

cilium/l7policy.h
```diff
@@ -18,8 +18,8 @@ namespace Cilium {
  * All Cilium L7 filter stats. @see stats_macros.h
  */
 // clang-format off
-#define ALL_CILIUM_STATS(COUNTER)                                                                  \
-  COUNTER(access_denied)                                                                                \
+#define ALL_CILIUM_STATS(COUNTER)	\
```
mhofstetter, line 21: there's still extra whitespace? is this necessary/formatted?

mhofstetter (commented):
nice :tada: splitting the two changes into two commits would have given extra points :smile:

mhofstetter (approved):

cilium/l7policy.h
jrajahalme: It is customary for the backslash within a macro definition to be some whitespace away, IMO.

Formatting if off for the macros (`// clang-format off` right above).

cilium/network_policy.cc
jrajahalme: I'd find it easier to check how these look in a sysdump (where I think we also dump all Envoy metrics?) and then document this, rather than try to trace the source code and figure out exactly how these come out :-)

mhofstetter: yeah, that's what i meant. Just gathering an example output from the Envoy metrics endpoint (yes we do this when fetching the sysdump)

PR #1004, opened 2024-11-09
tls_wrapper: Use raw socket with non-TLS policy

jrajahalme:
Cilium tls_wrapper should allow raw socket to be used if policy allows without TLS context.

Also check for SNI when getting TLS context so that the one matching the SNI is used if multiple are available.

Add policy tests validating the policy functionality.

Fixes: #890


jrajahalme:
Testing at https://github.com/cilium/cilium/pull/35887

sayboras (approved):
Nice coverage :heavy_check_mark:

PR #1020, opened 2024-12-02
Tests build all targets

jrajahalme:
Include also non-test targets in the `envoy-test-deps` target in hopes to make running tests with the tests-archive faster.

Other miscellaneous fixes:
- Add a new patch to fix segfault in integration tests when compiled for debug by testing for nullptr in ASSERT (ASSERT is a no-op in release builds)
- Update tools/stack_decode.py (even though it seems to require a GCC debug build to work)
- Update main branch Envoy version in README.md

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -25,7 +25,7 @@ build:release --fission=dbg
 
 # Manual link stamping, forces link to include current git SHA even if binary is otherwise
 # upto-date
-build --define manual_stamp=manual_stamp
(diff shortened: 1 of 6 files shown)


jrajahalme:
Integration tests took more than 10 minutes to build and run mainly because the tests archive did not have all the targets or their dependencies built. After this merges the test run on a PR that does not change any compiler options or source lines should not require any time for building, just running the tests.

sayboras (approved):

PR #1021, opened 2024-12-02
policy: Change abort to an error log

jrajahalme:
Log an error instead of crashing when Cilium NetworkPolicy resources are destructed in a worker thread. Remove stacktrace that is not useful at all with release builds.

Prior to enabling use of SDS secrets running NetworkPolicy destructors in a worker thread did not cause visible problems, even though we had taken measures to not do that a long time ago. Now that references to the policy have been removed from the connection metadata (Cilium SocketOption) this should no longer trigger. Nonetheless, crashing Envoy is too drastic as the event may be survivable (as it apparently has been for a long time when running without SDS references in the policy).

Enable trace level logging for troubleshooting if this error ever occurs.

Fixes: #1010

### DIFF
--- a/cilium/network_policy.cc
+++ b/cilium/network_policy.cc
@@ -608,12 +608,8 @@ public:
 
   ~PortNetworkPolicyRules() {
     if (!Thread::MainThread::isMainOrTestThread()) {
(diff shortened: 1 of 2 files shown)


sayboras (approved):
Thanks

PR #1057, opened 2024-12-13
policy: Remove node id from stats prefix

jrajahalme:
Simplify stats prefix so that this:

  cilium.policymap.127.0.0.1.3.policy.updates_rejected

becomes:

  cilium.policy.updates_rejected

### DIFF
--- a/cilium/network_policy.cc
+++ b/cilium/network_policy.cc
@@ -1102,24 +1102,21 @@ private:
 // This is used directly for testing with a file-based subscription
 NetworkPolicyMap::NetworkPolicyMap(Server::Configuration::FactoryContext& context)
     : context_(context.serverFactoryContext()), tls_map_(context_.threadLocal()),
-      local_ip_str_(context_.localInfo().address()->ip()->addressAsString()),
-      name_(fmt::format("cilium.policymap.{}.{}.", local_ip_str_, ++instance_id_)),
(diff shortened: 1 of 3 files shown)


sayboras (approved):

cilium/network_policy.cc
```diff
@@ -1104,17 +1104,16 @@ NetworkPolicyMap::NetworkPolicyMap(Server::Configuration::FactoryContext& contex
     : context_(context.serverFactoryContext()), tls_map_(context_.threadLocal()),
       local_ip_str_(context_.localInfo().address()->ip()->addressAsString()),
       name_(fmt::format("cilium.policymap.{}.{}.", local_ip_str_, ++instance_id_)),
```
mhofstetter, line 1106: can we  clean this up even further and remove the instance variables `name_` & `instance_id_`. 

Looks like now the name is only used in a log statement where it could be inlined (dynamic part (IP & ID) shouldn't be of relevance there too)

jrajahalme: instance id has been informative before, so I'll keep it

mhofstetter: as a result of removing `name_` we could also get rid of  the field `local_ip_str_` :thinking:

jrajahalme: Updated

mhofstetter (approved):

PR #1071, opened 2024-12-19
socket_options: small cleanup regarding setting socket options (IP_TRANSPARENT & SO_REUSEADDR)

mhofstetter:
```
socket_options: combine setting IP_TRANSPARENT for ipv4 & ipv6

This commit reduces boilerplate by reducing the different codepaths
for setting the IP_TRANSPARENT socket option for IPv4 & IPv6.
```

```
socket_options: move comment for SO_REUSEADDR

This commit moves the comment about setting SO_REUSEADDR
socket option to the code that is actually setting the option.
```

### DIFF
--- a/cilium/socket_option.h
+++ b/cilium/socket_option.h
@@ -96,48 +96,42 @@ public:
       // retries to not fail on "address already in use" when using a specific
       // source address and port.
 
-      // Set ip transparent option based on the socket address family
-      if (*ipVersion == Network::Address::IpVersion::v4) {
-        auto status = cilium_calls.setsockopt(socket.ioHandle().fdDoNotUse(), SOL_IP,
-                                              IP_TRANSPARENT, &one, sizeof(one));
-        if (status.return_value_ < 0) {
-          if (status.errno_ == EPERM) {
-            // Do not assert out in this case so that we can run tests without
-            // CAP_NET_ADMIN.
-            ENVOY_LOG(critical,
-                      "Failed to set socket option IP_TRANSPARENT, capability "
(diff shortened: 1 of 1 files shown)


cilium/socket_option.h
```diff
@@ -96,48 +96,40 @@ class SocketMarkOption : public Network::Socket::Option,
...
+        auto ip_transparent_socket_option_name = "IP_TRANSPARENT";
+        if (*ipVersion == Network::Address::IpVersion::v6) {
+          ip_transparent_socket_option = IPV6_TRANSPARENT;
+          ip_transparent_socket_option_name = "IPV6_TRANSPARENT";
         }
-      } else if (*ipVersion == Network::Address::IpVersion::v6) {
-        auto status = cilium_calls.setsockopt(socket.ioHandle().fdDoNotUse(), SOL_IPV6,
-                                              IPV6_TRANSPARENT, &one, sizeof(one));
+
+        auto status = cilium_calls.setsockopt(socket.ioHandle().fdDoNotUse(), SOL_IP,
```
jrajahalme, line 104: The use of `SOL_IP` for both IPv4 and IPv6 (instead of `SOL_IPV6`) seems suspect here?

jrajahalme (changes requested):
Should keep on using `SOL_IPV6` for IPv6.

cilium/socket_option.h
mhofstetter: good point. fixed

jrajahalme (approved):
Thanks for the clean-up :-)

PR #1072, opened 2024-12-20
bpf_metadata: cleanup  case for N/S L7 LB

mhofstetter:


sayboras (approved):

cilium/bpf_metadata.cc
```diff
@@ -406,38 +406,41 @@ Cilium::SocketOptionSharedPtr Config::getMetadata(Network::ConnectionSocket& soc
...
+      // IP family of the connection has no configured local ingress source address
+      ENVOY_LOG(
+          warn,
+          "cilium.bpf_metadata (north/south L7 LB): No local Ingress IP source address configured "
+          "for the family of {}",
+          sip->addressAsString());
       return nullptr;
     }
 
-    pod_ip = ip->addressAsString();
```
jrajahalme, line 421: IMO better not repeat `ingress_ip->addressAsString()` calls below. If you do not like setting `pod_ip` this early, how about:
```
auto& ingress_ip_str = ip->addressAsString();
```
and then later:
```
pod_ip = ingress_ip_str;
```

jrajahalme (changes requested):
small nit about repeating calls to `addressAsString()`

cilium/bpf_metadata.cc
mhofstetter: thanks, good point. fixed

cilium/bpf_metadata.cc
```diff
@@ -406,38 +406,42 @@ Cilium::SocketOptionSharedPtr Config::getMetadata(Network::ConnectionSocket& soc
...
-
-    if (!ip) {
-      // IP family of the connection has no configured local source address
-      ENVOY_LOG(warn,
-                "cilium.bpf_metadata (north/south L7 LB): No local IP source address configured "
-                "for the family of {}",
-                pod_ip);
+    // Pick the local ingress source address of the same family as the incoming connection
+    const Network::Address::Ip* ingress_ip = selectIPVersion(sip->version(), source_addresses);
+    auto& ingress_ip_str = ingress_ip->addressAsString();
```
jrajahalme, line 411: This would crash if `ingress_ip` is `nullptr`, tested above. So have to move after that test.

jrajahalme (changes requested):

cilium/bpf_metadata.cc
mhofstetter: glad that i have your second pair of eyes this morning :facepalm: thanks

jrajahalme (approved):

PR #1082, opened 2024-12-23
Policy destruct in main thread

jrajahalme:
Add an upstream patch to make sure the policy map is destructed from the main thread. The changed behavior triggers only when all listeners with Cilium `bpf_metadata` filter are removed, and the listener drain time (default 10 minutes) passes.


sayboras (approved):
Seems good locally and CI 👍

PR #1086, opened 2025-01-03
gitignore: disallow dot files by default and enable explictly

mhofstetter:
Currently, dotfiles are explictly git-ignored via `.gitignore` file.

Instead of adding additional ignores (e.g. direnv' `.envrc`), this commit adopts the upstream pattern of disallowing all dot files by default, and enable explictly.

See: https://github.com/envoyproxy/envoy/blob/main/.gitignore#L1


sayboras (approved):
:ship:

PR #1099, opened 2025-01-06
bazel: register c++ toolchain via build flag --extra_toolchains

mhofstetter:
Currently, the C++ toolchains (defined in `bazel/toolchains`) are
registered in the `WORKSPACE` file via `register_toolchains`.

Problem with this approach is that this is tightly coupled to the Cilium Proxy
CI environment (Ubuntu & expected paths in `/usr/bin/...`). If these requirements
aren't met, compilation with the evaluated C++ toolchain  fails - even though 
there might be an available and compatible local toolchain.

To provide more flexibility in terms of development environment and integration,
this commit removes the hardcoded registration of the toolchains in the `WORKSPACE`
file and replaces it with an explicit registration via bazel build argument
`--extra_toolchains="//bazel/toolchains:all"` where required (e.g. CI).

Example: Using a downloaded and extracted Clang&LLVM as [documented in upstream Envoy](https://github.com/envoyproxy/envoy/blob/main/bazel/README.md#linux) doesn't work as expected due to the hardcoded tool  paths in the defined Cilium Proxy C++ toolchains. Hence it would be nice to rely on the local toolchain in this case.

### DIFF
--- a/.github/workflows/build-envoy-image-ci.yaml
+++ b/.github/workflows/build-envoy-image-ci.yaml
@@ -89,7 +89,7 @@ jobs:
           build-args: |
             BUILDER_BASE=quay.io/cilium/cilium-envoy-builder-dev:${{ env.BUILDER_DOCKER_HASH }}
             ARCHIVE_IMAGE=quay.io/${{ github.repository_owner }}/cilium-envoy-builder:main-archive-latest
-            BAZEL_BUILD_OPTS=--remote_upload_local_results=false
+            BAZEL_BUILD_OPTS=--remote_upload_local_results=false --extra_toolchains=//bazel/toolchains:all
           cache-from: type=local,src=/tmp/buildx-cache
           cache-to: type=local,dest=/tmp/buildx-cache,mode=max
           push: true
--- a/.github/workflows/build-envoy-image-ci.yaml
+++ b/.github/workflows/build-envoy-image-ci.yaml
@@ -1,6 +1,6 @@
 name: CI Build & Push
 on:
-  pull_request_target:
+  pull_request:
     types: [opened, synchronize, reopened]
 
 permissions:
--- a/.github/workflows/build-envoy-images-release.yaml
+++ b/.github/workflows/build-envoy-images-release.yaml
@@ -78,7 +78,7 @@ jobs:
             BUILDER_BASE=quay.io/${{ github.repository_owner }}/cilium-envoy-builder:test-${{ env.BUILDER_DOCKER_HASH }}
             ARCHIVE_IMAGE=quay.io/${{ github.repository_owner }}/cilium-envoy-builder:test-main-archive-latest
             COPY_CACHE_EXT=.new
-            BAZEL_BUILD_OPTS="--jobs=HOST_CPUS*.75"
+            BAZEL_BUILD_OPTS=--jobs=HOST_CPUS*.75 --extra_toolchains=//bazel/toolchains:all
             BAZEL_TEST_OPTS=--test_timeout=300 --local_test_jobs=1 --flaky_test_attempts=3
           push: true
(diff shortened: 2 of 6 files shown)


mhofstetter:
@jrajahalme @sayboras what do you think about this change? would this still be compatible with your dev workflow? (maybe by having to set the bazel build argument too? (unfortunately there's no command to disable automatic C++ toolchain evaluation - at least to my knowledge)).

Without this change, I either have to remove the toolchain registration in `WORKSPACE` or change all tool paths in the toolchain definition.

Note: Yes, i have to revert the github action changes (`pull_request` -> `pull_request_target`)

sayboras:
> what do you think about this change? would this still be compatible with your dev workflow?

I was actually disabling this in my local dev for a while unless I am using ubuntu box, so there is no impact from my side. AFAIK, wofi build is also doing the same.

https://github.com/wolfi-dev/os/blob/567e2634fd9062e955535dd024c602283f554a44/cilium-envoy-1.16.yaml#L97

jrajahalme (changes requested):
All make targets that use CI produced archives need to also specify the `--extra_toolchains` so that the builds use the same toolchain as was used to build those images.

`make docker-tests` is the easiest thing to check for. On `main` it should pull the CI archive and pretty much just run the tests without compiling anything, except for any files changed locally. But there are others to look for as well.

mhofstetter:
> how about somehow makefiles detecting if the defined toolchains are functional and setting the `--extra_toolchains` if so, and leaving it out otherwise, maybe with a message that a local toolchain is used and must be available for compilation to succeed?

good idea. would it be an idea to use `PKG_BUILD` in `Makefile` (https://github.com/cilium/proxy/blob/main/Makefile#L79) as an indicator to set ``--extra_toolchains` to `BAZEL_BUILD_OPTS`? :thinking: AFAIU/K this should cover all cases that are building within the CI docker environment? :thinking:

jrajahalme:
> > how about somehow makefiles detecting if the defined toolchains are functional and setting the `--extra_toolchains` if so, and leaving it out otherwise, maybe with a message that a local toolchain is used and must be available for compilation to succeed?
> 
> good idea. would it be an idea to use `PKG_BUILD` in `Makefile` (https://github.com/cilium/proxy/blob/main/Makefile#L79) as an indicator to add `--extra_toolchains=...` to `BAZEL_BUILD_OPTS`? 🤔 AFAIU/K this should cover all cases that are building within the CI docker environment (because these are the ones setting `PKG_BUILD`)? 🤔

This would be cleaner than adding `--extra-toolchans` everywhere! Local make targets still need it though (e.g., `docker-tests`).

mhofstetter:
> > > how about somehow makefiles detecting if the defined toolchains are functional and setting the `--extra_toolchains` if so, and leaving it out otherwise, maybe with a message that a local toolchain is used and must be available for compilation to succeed?
> > 
> > 
> > good idea. would it be an idea to use `PKG_BUILD` in `Makefile` (https://github.com/cilium/proxy/blob/main/Makefile#L79) as an indicator to add `--extra_toolchains=...` to `BAZEL_BUILD_OPTS`? 🤔 AFAIU/K this should cover all cases that are building within the CI docker environment (because these are the ones setting `PKG_BUILD`)? 🤔
> 
> This would be cleaner than adding `--extra-toolchans` everywhere! Local make targets still need it though (e.g., `docker-tests`).

AFAIK it's just that the "outer" make call doesn't have `PKG_BUILD` set and therefore doesn't uses the C++ toolchain. But it doesn't execute bazel and therefore it's ok.

Because it's the "inner" make call that gets executed in docker via `Dockerfile.tests` that has `PKG_BUILD` set to `1` and therefore uses the correct toolchain.

except if we would miss some `PKG_BUILD=1` on an "inner" call

I'll close this PR in favor of https://github.com/cilium/proxy/pull/1101 (undrafted) - let's continue any potential discussion over there.

PR #1087, opened 2025-01-03
build: improve clangd LSP server support

mhofstetter:
See the individual commits for more information

* build: add tools/gen_compilation_database.py
* clangd: add clangd configuration

Wait for dependent PR to be merged: https://github.com/cilium/proxy/pull/1086 (edit: :heavy_check_mark:)

PR #1100, opened 2025-01-07
code: cleanup includes - use "include what you use" model

mhofstetter:
Currently, the C++ files are either including unused files or heavily rely on transtive includes and doesn't define the actual required includes.

Therefore, this commit fixes all includes so that all files are defining their includes according to their needs - following the model "include what you use".

See https://clangd.llvm.org/guides/include-cleaner

Advantages:

* Get rid of warnings when using clangd as LSP server
* Remove unused includes
* Define all direct includes helps understanding the code and the dependencies

Note: For the time being there will not be any validation in CI. It's more or less just a cleanup and a potential start to establish this pattern because I think this helps for others to familiarize with the codebase.

Related to https://github.com/cilium/proxy/pull/1087

### DIFF
--- a/cilium/accesslog.cc
+++ b/cilium/accesslog.cc
@@ -1,13 +1,32 @@
 #include "accesslog.h"
 
-#include <errno.h>
 #include <stdlib.h>
 #include <sys/socket.h>
 #include <sys/un.h>
 #include <unistd.h>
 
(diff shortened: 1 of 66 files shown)


jrajahalme (approved):
Make clangd happy :-)

PR #1107, opened 2025-01-09
socket_options: split into multiple socket options

mhofstetter:
Split the existing 2 socket options into more fine-grain and cohesive socket options

* IP_TRANSPARENT
* SO_REUSEADDR
* SO_REUSEPORT
* SO_MARK (Cilium Mark)
* Cilium BPF Metadata
* Cilium Source Address & Identity

This split has the following advantages:

* Separate configuration of the individual socket options for listener- and connection sockets (instead of doing an in-socket-option decision based on the security id (`0` = listener socket))
* Scope and responsibility of the socket options is clear easier to understand (currently the all-in-one socket option is responsible to pass information to filters (via filter state), pass information to Cilium network- and http filters (via Envoy socket option functionality), configure the upstream connection pool separation, setting actual options on the socket, restoring the socket's local address)


cilium/socket_option_reuse_addr.h
```diff
@@ -0,0 +1,68 @@
...
+#include "source/common/common/logger.h"
+#include "source/common/common/utility.h"
+
+#include "absl/types/optional.h"
+
+namespace Envoy {
+namespace Cilium {
+
+// Socket Option that programmatically sets the socket option SO_REUSEADDR on the socket.
+class ReuseAddrSocketOption : public Network::Socket::Option,
```
mhofstetter, line 22: maybe we can get rid of the explicit SocketOptions for `SO_REUSEADDR` & `SO_REUSEPORT` completely and use the upstream `SocketOptionImpl` (https://github.com/envoyproxy/envoy/blob/main/source/common/network/socket_option_impl.cc).

One thing that we would probably loose is some more logging (in case of successfully setting the socket option)

cilium/socket_option_bpf_metadata.h
```diff
@@ -0,0 +1,127 @@
...
+public:
+  virtual ~PolicyResolver() = default;
+
+  virtual uint32_t resolvePolicyId(const Network::Address::Ip*) const PURE;
+  virtual const PolicyInstanceConstSharedPtr getPolicy(const std::string&) const PURE;
+};
+
+// Socket Option that holds relevant connection & policy information that can be retrieved
+// by the Cilium network- and HTTP policy filters by using GetBpfMetadataSocketOption.
+class BpfMetadataSocketOption : public Network::Socket::Option,
```
mhofstetter, line 38: maybe worth to rename this to `PolicyMetadataSocketOption` or `PolicyContextSocketOption` - as it's used by the policy filters (network & http)?

cilium/socket_option_reuse_addr.h
mhofstetter: we could then also use the same approach to set the other socket options that we set in the bpf metadata listener filter.

* `SO_LINGER`
* `SO_KEEPALIVE`
* `TCP_KEEPINTVL`
* `TCP_KEEPIDLE`

cilium/bpf_metadata.cc
```diff
@@ -349,7 +358,8 @@ const PolicyInstanceConstSharedPtr Config::getPolicy(const std::string& pod_ip)
   return policy;
 }
 
-Cilium::SocketOptionSharedPtr Config::getMetadata(Network::ConnectionSocket& socket) {
+Cilium::BpfMetadata::SocketInformationSharedPtr
+Config::extractSocketInformation(Network::ConnectionSocket& socket) {
```
mhofstetter, line 362: Currently, extraction of the socket information is more or left as is. Just being renamed and returning an "intermediate" struct holding the required data that is then used to create the actual socket options.

It also still performs some actions on the socket itself.

* restoration of the local address (for the original dst cluster)
* setting the requested application protocol that is configured with proxylib

IMO extracting information shouldn't perform any actions on the socket at all. Probably worth to clean this up too (and move outside of this function) - maybe in a follow up PR.

cilium/socket_option_source_address.h
```diff
@@ -0,0 +1,146 @@
...
+    const uint8_t* byte_array = reinterpret_cast<const uint8_t*>(&address);
+    vec.insert(vec.end(), byte_array, byte_array + sizeof(T));
+  }
+
+  void hashKey(std::vector<uint8_t>& key) const override {
+    // Source address is more specific than policy ID. If using an original
+    // source address, we do not need to also add the source security ID to the
+    // hash key. Note that since the identity is 3 bytes it will not collide
+    // with neither an IPv4 nor IPv6 address.
+    if (original_source_address_) {
```
mhofstetter, line 102: Currently, the `hashKey` implementation only considers the original source address - but not the configurable ipv4/ipv6 source addresses that are passed in most of the cases.

I wonder whether this is correct - or whether this was missed when the ingress source addresses have been introduced. maybe we should use the source addresses for building the hash in these cases too (instead of falling back to use the source identity.

cilium/socket_option_reuse_addr.h
jrajahalme: Upstream `ListenerImpl::buildListenSocketOptions()` is doing this based on the listener config. We can not use those for any socket options that require privileges, though. But for any non-privileged options we could use those either from our code, or via listener config.

Listener config can also add options Envoy does not know about using numeric values for the options name etc. (`buildLiteralOptions()`)

There are also these:
- `Network::SocketOptionFactory::buildZeroSoLingerOptions()`
-  `SocketOptionFactory::buildTcpKeepaliveOptions(Network::TcpKeepaliveConfig keepalive_config)`
- (general pattern in Envoy source: "build.*Options")

Any options we need for CIlium datapath integration (all of them?) should be set from our code, rather than from listener config. We could add those in to CEC listeners as we are parsing them, though.

cilium/bpf_metadata.cc
jrajahalme: I like the term `metadata` vs. `information`, as it is derived from Cilium bpf datapath resources associated with the socket, rather than from the socket itself! So if you want to rename this, how about `extractSocketMetadata`?

cilium/socket_option_bpf_metadata.h
mhofstetter: `CiliumPolicySocketOption` :+1: (prefer to have a consistent suffix `SocketOption`)

cilium/bpf_metadata.cc
mhofstetter: makes sense. will rename the function & intermediate struct accordingly

cilium/socket_option_source_address.h
jrajahalme: Could be an issue, but when L7LB is used (E/W or N/S), then the original source address is not kept: https://github.com/cilium/proxy/blob/main/cilium/bpf_metadata.cc#L427

So the connection pool (hashKey) is restricted on the source identity only, which likely is WORLD etc.

cilium/bpf_metadata.cc
mhofstetter: renamed

cilium/socket_option_bpf_metadata.h
mhofstetter: renamed

cilium/socket_option_reuse_addr.h
mhofstetter: yep, even better if we can use the upstream builder functions.

pushed an updated version that illustrates the possibilities:

* `SO_REUSEPORT` -> use `buildReusePortOptions`
* `SO_REUSEADDR` -> `buildReuseAddrOptions` doesn't exist -> use raw `Envoy::Network::SocketOptionImpl` - still possible to delete our custom implementation
* `SO_LINGER` -> use `buildZeroSoLingerOptions` -> unfortunately this doesn't provide any options to configure the linger time - it's hardcoded to `0` (and we currently use `10`) -> either we change this or we have to use the raw `SocketOptionImpl` too
* `SO_KEEPALIVE` and tcp -> use `buildTcpKeepaliveOptions` that provides support to configure the timeouts. only change is that the current implementation uses best-effort when applying these settings - whereas the upstream implementation would fail if applying the socket options fail.

it's nice to see that we could get rid of quite some boilerplate when using the upstream code.

mhofstetter:
Closing in favor of https://github.com/cilium/proxy/pull/1115

PR #1113, opened 2025-01-13
policy: Remove initialization dependency from listener initialization

jrajahalme:
Listeners using Cilium network policy enforcement are shared among all the pods in the node, so the listener initialization should not stall on policy update at all, as this prevents the listener from actually listening on it's listen port before the initial network policy update has completed.

Moreover, a network policy update can stall for a long time if the required SDS secrets are missing. Reduce the timeout for fetching secrets to 50ms (from 5 seconds) to not stall policy updates on the node due to a missing secret, as the effect of the missing secret is typically limited to a subset of pods, while stalling policy updates has negative effect on all pods using `toFQDNs` policies, for example.

Break up Envoy init manager functionality into helpers to make the business logic easier to read.

Needs backport to v1.31 for Cilium v1.17 SDS support.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -20,8 +20,8 @@ build:release -c opt
 # No debug info for release builds
 build:release --define no_debug_info=1
 build:release --linkopt=-Wl,--strip-all
-build:release --features=-per_object_debug_info
-build:release --fission=dbg
+build --features=-per_object_debug_info
+build --fission=dbg
 
 # Manual link stamping, forces link to include current git SHA even if binary is otherwise
 # upto-date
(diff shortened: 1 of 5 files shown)


jrajahalme:
Fixed format.

mhofstetter (approved):
Change itself looks reasonable - even though I can't review the NPDS code with 100% confidence. (init manager etc.)

jrajahalme:
> Change itself looks reasonable - even though I can't review the NPDS code with 100% confidence. (init manager etc.)

Thanks for the review, let's see how this does in cilium/cilium...

PR #1115, opened 2025-01-14
bpf_metadata: cleanup socketoptions

mhofstetter:
This PR split the existing 2 socket options (SocketMarkOption & SocketOption (inherits from the first one)) into more fine-grain and cohesive socket options. In addition, upstream Envoy socket options are used where possible.

Please see the individual commits for details.

This split has the following advantages:

* Separate configuration of the individual socket options for listener- and connection sockets (instead of doing an in-socket-option decision based on the security id (0 = listener socket))
* Scope and responsibility of the socket options is clear and easier to understand (currently the all-in-one socket option is responsible to pass information to filters (via filter state), pass information to Cilium network- and http filters (via Envoy socket option functionality), configure the upstream connection pool separation, setting actual options on the socket, restoring the socket's local address)
* Remove the additional complexity due to the inheritance of the socket options
* Proper names for the socket options (as the global one added more and more functionality that we're no longer aligned with the name)

**Note: A follow up PR will convert the `CiliumPolicySocketOption` into a Envoy filter state object. https://github.com/cilium/proxy/pull/1116**

### DIFF
--- a/cilium/bpf_metadata.cc
+++ b/cilium/bpf_metadata.cc
@@ -47,6 +47,7 @@
 #include "cilium/network_policy.h"
 #include "cilium/policy_id.h"
 #include "cilium/socket_option.h"
+#include "cilium/socket_option_ip_transparent.h"
 
 namespace Envoy {
 namespace Server {
@@ -77,6 +78,8 @@ public:
 
     uint32_t mark = (config->is_ingress_) ? 0x0A00 : 0x0B00;
     options->push_back(std::make_shared<Cilium::SocketMarkOption>(mark, 0));
+
+    options->push_back(std::make_shared<Cilium::IpTransparentSocketOption>());
 
     options->push_back(std::make_shared<Envoy::Network::SocketOptionImpl>(
         envoy::config::core::v3::SocketOption::STATE_PREBIND,
@@ -131,6 +134,8 @@ public:
 
     uint32_t mark = (config->is_ingress_) ? 0x0A00 : 0x0B00;
     options->push_back(std::make_shared<Cilium::SocketMarkOption>(mark, 0));
+
+    options->push_back(std::make_shared<Cilium::IpTransparentSocketOption>());
 
     options->push_back(std::make_shared<Envoy::Network::SocketOptionImpl>(
         envoy::config::core::v3::SocketOption::STATE_PREBIND,
@@ -580,6 +585,10 @@ Network::FilterStatus Instance::onAccept(Network::ListenerFilterCallbacks& cb) {
               Network::UpstreamSocketOptionsFilterState::key())
           ->addOption(std::move(options));
     }
+  }
+
+  if (config_->addPrivilegedSocketOptions()) {
+    socket_options->push_back(std::make_shared<Envoy::Cilium::IpTransparentSocketOption>());
   }
 
   // allow reuse of the original source address by setting SO_REUSEADDR.
--- a/cilium/bpf_metadata.cc
+++ b/cilium/bpf_metadata.cc
@@ -349,7 +349,8 @@ const PolicyInstanceConstSharedPtr Config::getPolicy(const std::string& pod_ip) 
   return policy;
 }
 
-Cilium::SocketOptionSharedPtr Config::getMetadata(Network::ConnectionSocket& socket) {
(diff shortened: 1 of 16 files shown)


cilium/bpf_metadata.cc
```diff
@@ -546,7 +584,7 @@ Network::FilterStatus Instance::onAccept(Network::ListenerFilterCallbacks& cb) {
       }
 
       auto options = std::make_shared<Network::Socket::Options>();
-      options->push_back(std::move(option));
+      options->push_back(std::move(cilium_policy_socket_option));
```
mhofstetter, line 587: note: only the `CiliumPolicySocketOption` is set in the filterstate `UpstreamSocketOptionsFilterState` - and therefore passed to the upstream sockets.

AFAIU this is ok - and we don't have to add all the other options. Reasoning for this is that this functionality has been added with https://github.com/cilium/proxy/pull/581 - and the only reason mentioned is that the policy information need to be potentially available for upstream filters.

cilium/bpf_metadata.cc
```diff
@@ -349,7 +349,8 @@ const PolicyInstanceConstSharedPtr Config::getPolicy(const std::string& pod_ip)
   return policy;
 }
 
-Cilium::SocketOptionSharedPtr Config::getMetadata(Network::ConnectionSocket& socket) {
+Cilium::BpfMetadata::SocketMetadataSharedPtr
+Config::extractSocketMetadata(Network::ConnectionSocket& socket) {
```
jrajahalme, line 376: The intermediate result should not need to be a (shared) pointer, should use `absl::optional` instead. Upstream code base has plenty of examples for it :-)

cilium/socket_option_ip_transparent.h
```diff
@@ -0,0 +1,93 @@
...
+
+    // Set ip transparent option based on the socket address family
+    auto ip_transparent_socket_option = IP_TRANSPARENT;
+    auto ip_transparent_socket_option_name = "IP_TRANSPARENT";
+    if (*ipVersion == Network::Address::IpVersion::v6) {
+      ip_transparent_socket_option = IPV6_TRANSPARENT;
+      ip_transparent_socket_option_name = "IPV6_TRANSPARENT";
+    }
+
+    auto status = cilium_calls.setsockopt(socket.ioHandle().fdDoNotUse(), SOL_IP,
```
jrajahalme, line 64: Should this be `SOL_IPV6` for IPv6?

cilium/bpf_metadata.cc
```diff
@@ -132,6 +135,8 @@ class UdpBpfMetadataConfigFactory : public NamedUdpListenerFilterConfigFactory {
     uint32_t mark = (config->is_ingress_) ? 0x0A00 : 0x0B00;
     options->push_back(std::make_shared<Cilium::SocketMarkOption>(mark, 0));
 
+    options->push_back(std::make_shared<Cilium::IpTransparentSocketOption>());
```
jrajahalme, line 138: should this be conditional on`addPrivilegedSocketOptions` as well?

cilium/bpf_metadata.cc
```diff
@@ -78,6 +79,8 @@ class BpfMetadataConfigFactory : public NamedListenerFilterConfigFactory {
...
```
jrajahalme, line 82: should this be conditional on`addPrivilegedSocketOptions` as well?

jrajahalme (changes requested):
Thanks, nice work! I'd like you to check the nits below before we merge!

cilium/bpf_metadata.cc
mhofstetter: thanks for the hint - refactored

cilium/socket_option_ip_transparent.h
mhofstetter: of course - even though this was still "fixed" on `main` it was still wrong on my working copy where i copied this from. thanks a lot!

cilium/bpf_metadata.cc
mhofstetter: this code path isn't executed during tests - hence no need to add the condition.

cilium/bpf_metadata.cc
mhofstetter: this code path isn't executed during tests - hence no need to add the condition.

PR #1116, opened 2025-01-14
bpf_metadata: convert policy socket option into filter state

mhofstetter:
PR #1115 splitted the existing Cilium socket options into multiple smaller ones. 

But the Socket Option `CiliumPolicySocketOption` doesn't actually set any "socket options" on the network socket. It just uses Envoy's Socket Option facility to pass the policy related information to the filters and upstream connection.

In the meantime, upstream Envoy introduced the concept of filterstate. Therefore, this PR converts the socket option `CiliumPolicySocketOption` into a filter state.

Follow up of https://github.com/cilium/proxy/pull/1115


jrajahalme (approved):
Thanks! This been long time wanted :-)

Did not recall we were already using filter state to give upstream filters access to the policy!

The last commit underlines why this is so nice, cleaning up the policy users who no longer need to depend on socket related library.

jrajahalme:
Labeled this and #1115 for 1.31 backport so that #1114 can be backported without conflicts.

PR #1119, opened 2025-01-16
bpf_metadata: extract proxylib appprotocol logic

mhofstetter:
Currently, the function `extractSocketMetadata` does not only extract relevant information from the socket and enriches it with data retrieved from BPF maps. It also has some side-effects: namely setting the proxylib protocol as application protocol on the network socket.

Therefore, this commit extracts this logic from the function `extractSocketMetadata` and moves it into a dedicated function on the `SocketMetadata` struct that then is called from the BPF  metadata listener filter in the `onAccept` function.

This way, we can keep the function `extractSocketMetadata` free of side-effects and bundle the logic that changes the socket in the BPF metadata listener filter (`onAccept`).

Note: `socket.connectionInfoProvider().restoreLocalAddress(dst_address)` - the last modifying action in `extractSocketMetadata`) will be cleaned up in a follow up PR.

As discussed over [here](https://github.com/cilium/proxy/pull/1107#discussion_r1909271347)

### DIFF
--- a/cilium/bpf_metadata.cc
+++ b/cilium/bpf_metadata.cc
@@ -6,7 +6,6 @@
 #include <netinet/tcp.h>
 #include <sys/socket.h>
 
-#include <chrono>
 #include <cstddef>
 #include <cstdint>
 #include <memory>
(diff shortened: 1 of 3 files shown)


mhofstetter:
Closing in favor of https://github.com/cilium/proxy/pull/1129 (using the same policy to extract the proxylib l7 proto)

PR #1131, opened 2025-01-20
bpf_metadata: add unit-test for proxylib l7 proto

mhofstetter:
This commit adds an unit test that verifies that the proxylib l7 protocol has been correctly set to the request application protocols of the network socket.

### DIFF
--- a/tests/metadata_config_test.cc
+++ b/tests/metadata_config_test.cc
@@ -5,6 +5,7 @@
 #include <cstdint>
 #include <list>
 #include <memory>
+#include <string>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


tests/metadata_config_test.cc
```diff
@@ -481,6 +482,40 @@ TEST_F(MetadataConfigTest, EastWestL7LbMetadataNoOriginalSource) {
...
+  ::cilium::BpfMetadata config{};
+
+  EXPECT_NO_THROW(initialize(config));
+
+  auto socket_metadata = config_->extractSocketMetadata(socket_);
+  EXPECT_TRUE(socket_metadata);
+
+  socket_metadata->configureProxyLibApplicationProtocol(socket_);
+
+  EXPECT_CALL(socket_, setRequestedApplicationProtocols(_)).Times(0);
```
jrajahalme, line 495: Not sure if absolutely needed, but it would be cleaner if `EXPECT_CALL` would be before the call above, like in the other test!

jrajahalme (changes requested):
One nit

tests/metadata_config_test.cc
mhofstetter: changed - thanks!

jrajahalme (approved):
Thanks for the tests!

PR #1133, opened 2025-01-21
bpf_metadata: extract restoreLocalAddress

mhofstetter:
Currently, `extractSocketMetadata` also restores the local address for the use by the `OriginalDestCluster`.

To keep the function free of side-effects, the logic that restores the local address gets extracted into a function within the `SocketMetadata` that gets called from the BPF metadata listener after extracting the BPF metadata from the socket.


jrajahalme (approved):

jrajahalme (approved):

sayboras (approved):
Nice, also the test 👍

PR #1143, opened 2025-01-24
bpf_metadata: align logging for restoration

mhofstetter:
This commit aligns the logging for restoration of the local address for downstream & upstream socket by always logging the socket and the old & new local address.

### DIFF
--- a/cilium/socket_option_source_address.h
+++ b/cilium/socket_option_source_address.h
@@ -82,11 +82,16 @@ public:
       return true;
     }
 
-    // Note: we don't expect this to change the behaviour of the socket. We expect it to be copied
-    // into the upstream connection later.
+    // Note: Restoration of the original source address happens on the socket of the upstream
+    // connection.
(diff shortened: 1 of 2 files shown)


cilium/socket_option_source_address.h
```diff
@@ -84,9 +84,14 @@ class SourceAddressSocketOption : public Network::Socket::Option,
 
     // Note: we don't expect this to change the behaviour of the socket. We expect it to be copied
     // into the upstream connection later.
```
jrajahalme, line 86: Is this comment still valid?

jrajahalme (approved):
One question on a validity of a comment.

cilium/socket_option_source_address.h
mhofstetter: still valid - but the whole `setOption` is executed on the upstream socket.

I slightly adjusted the comment.

jrajahalme (approved):

PR #1151, opened 2025-01-28
bpf_metadata: set TCP keep alive socket options without Envoy SocketOption

mhofstetter:
PR https://github.com/cilium/proxy/pull/1115 unintentionally changed that the keep alive socket options are no longer applied on the accepted downstream socket, instead (due to the usage of Envoy SocketOption) they are now applied on the corresponding upstream socket.

This commit is reverting this change and adds some helping comments to the code.

Note: Same would have applied for `SO_LINGER` - but this has been removed completely in the meantime with https://github.com/cilium/proxy/pull/1128

### DIFF
--- a/cilium/bpf_metadata.cc
+++ b/cilium/bpf_metadata.cc
@@ -28,6 +28,7 @@
 #include "envoy/stream_info/filter_state.h"
 
 #include "source/common/common/logger.h"
+#include "source/common/common/utility.h"
 #include "source/common/network/address_impl.h"
 #include "source/common/network/socket_option_factory.h"
 #include "source/common/network/socket_option_impl.h"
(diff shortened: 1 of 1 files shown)


cilium/bpf_metadata.cc
```diff
@@ -561,17 +567,36 @@ Network::FilterStatus Instance::onAccept(Network::ListenerFilterCallbacks& cb) {
...
-          .keepalive_interval_ = 5 * 60,      // 5 min
-      }));
-
+  // Adding SocketOptions to the downstream socket. The function `setOption` is NOT executed
+  // on the downstream socket itself - it's executed later on the corresponding upstream socket!
   socket.addOptions(socket_options);
 
+  // set keep alive socket options on accepted connection socket
+  // (SO_KEEPALIVE, TCP_KEEPINTVL, TCP_KEEPIDLE)
+  int keepalive = true;
```
mhofstetter, line 576: see original commit: https://github.com/cilium/proxy/pull/1115/commits/13d6409dbc1e168a002cad87c4e2c6a41c38f3f7

it's not a 1:1 revert - error handling is different and fails in case socket option can't be applied.

jrajahalme (approved):

cilium/bpf_metadata.cc
jrajahalme: This is fine :-)

PR #1166, opened 2025-02-10
envoy: skip upstream handling on shadows

mhofstetter:
PR https://github.com/cilium/proxy/pull/1140 introduced a regression where
logging fails on shadow upstream connections (Detected in Gateway API tests in cilium/cilium).

Therefore this PR is checking for shadowing connections and skips
enforcement & logging on them.


sayboras (approved):
Looks good to me, considered that we have done similar thing in https://github.com/cilium/proxy/pull/581

PR #1172, opened 2025-02-18
policy: Use source pod's egress policy if available

jrajahalme:
Enforce source pod's egress policy, if available, i.e., when the source is a local pod, even in the north/south scenario, where the Ingress IP is used as the upstream source address.


mhofstetter (approved):

jrajahalme:
Builder image appears broken, apt is crashing on libc update. Adding clang-tidy to to the builder image to cause a new one be built & we need it soon anyway.

sayboras (approved):
Touch base offline for some points I don't understand, thanks ✅

jrajahalme:
Removed stray comment.

PR #1175, opened 2025-02-19
envoy: Bump envoy to v1.33.0

sayboras:
This is as part of regular maintenance work, please refer to individual commit for more details.

Just a note that 1.31 will be EOL in 3 months, so we will maintain two branches:
- v1.32 will be used for all cilium/cilium releases
- v1.33 will be for development with cilium/cilium main branch
   - This will also pave the way for ubuntu 24.04 upgrade as well.


jrajahalme (approved):
Awesome work, thanks!

PR #1178, opened 2025-02-22
gha: Use workflow_call for Cilium Integration test

sayboras:
The goal is to avoid extra waiting time, while building the images.

Testing was done with temp commit, here is the sample run https://github.com/cilium/proxy/actions/runs/15178091834/job/42682174454?pr=1178


viktor-kurchenko (approved):
I like it!

PR #1190, opened 2025-02-26
ipcache: Try multiple maps of different sizes

jrajahalme:
Try open `cilium_ipcache_v2` and `cilium_ipcache`, in this order.

Only use the first 4 bytes of the map value.

### DIFF
--- a/cilium/bpf.cc
+++ b/cilium/bpf.cc
@@ -21,15 +21,22 @@ enum {
   BPF_KEY_MAX_LEN = 64,
 };
 
-Bpf::Bpf(uint32_t map_type, uint32_t key_size, uint32_t value_size)
-    : fd_(-1), map_type_(map_type), key_size_(key_size), value_size_(value_size) {}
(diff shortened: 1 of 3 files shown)


pchaigno (approved):
I didn't review the code, but I tested it in https://github.com/cilium/cilium/pull/36359 (which has the new ipcache v2 logic) and I can confirm it works. It fixes L7 tests that were previously failing.

pchaigno:
@jrajahalme @sayboras I think this made it into cilium-proxy v1.32, which then made it into Cilium v1.17 and broke downgrades in my IPv6 tunneling PR. On upgrade from v1.17, we want to try the v2 ipcache first, then try the v1. But on downgrade to v1.17, I think we only want to open v2 if we can't find v1. Otherwise, we risk getting stale security identities from the v2 map.

To fix that, can I simply send a patch to the v1.32 branch here?

sayboras:
> can I simply send a patch to the v1.32 branch here?

Yes, please go ahead :+1:

pchaigno:
I had not even closed the tab that I already have an answer :exploding_head: 

:pray: Tam!

PR #1195, opened 2025-03-06
WIP: use cncf hosted gha runners

jeefy:
CNCF has hosted ephemeral GitHub runners in Oracle that we're wanting projects to use rather than the GitHub hosted ones, which are now incur a cost to use. This PR is currently a WIP to work through any tests that break or dependencies that may be missing. <3

Please direct any questions to myself, @krook and @RobertKielty

### DIFF
--- a/.github/workflows/build-envoy-image-ci.yaml
+++ b/.github/workflows/build-envoy-image-ci.yaml
@@ -17,7 +17,7 @@ jobs:
   build-and-push-prs:
     timeout-minutes: 360
     name: Build and push multi-arch images
-    runs-on: ubuntu-latest-64-cores-256gb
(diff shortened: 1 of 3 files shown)


jeefy:
Somewhat consistent test failure here: https://github.com/cilium/proxy/actions/runs/13709538606/job/38386691311#step:17:1670

Immediate thinking is it's something with the job running in a pod and not a VM (primary difference between GH hosted and self-hosted using GitHub ARC Runners) -- Ideas appreciated before I dive down this rabbit hole :)

jeefy:
@sayboras Any idea on why this would still be running on the hosted GH runner rather than the CNCF Runner? https://github.com/cilium/proxy/actions/runs/13729868767/job/38404564433?pr=1195#step:1:2

PR #1208, opened 2025-03-14
proxylib-go: cleanup go module

mhofstetter:
This PR does a cleanup of the proxylib go module. Historically this is just a copy of some `cilium/cilium` packages that were [copied over when moving the proxylib go code from](https://github.com/cilium/proxy/pull/232) `cilium/cilium` into `cililum/proxy`.

* Remove unused / unnecessary functionality in packages `pkg/...`
* Remove unused thirdparty dependencies that are no longer used after removing the unnecessary functionality (Reduce binary size, reduce renovate churn, reduce potential transitive dependencies into `cilium/cilium`)
* Reduce API surface of `cilium/proxy` -> Now it only exposes the Kafka L7 proto types which are actually used by `cilium/cilium` (`pkg/policy/api/kafka`) (of course aside from the rest of the xDS types).

Please review the individual commits.

Follow ups:
* move back `pkg/policy/api/kafka` to `cilium/cilium` (as it's only shared for the kafka api key constants...) 
* logrus -> slog migration
* remove checkmate


sayboras (approved):
This is awesome 😲

PR #1219, opened 2025-03-19
gha: Remove un-used command/env/file

sayboras:
This variable is not used in github action workflow, so better just remove it.

Makefile derives the value from git command.


mhofstetter:
Can you please provide some more context why this is not needed (PR desc or commit message)

mhofstetter (approved):

PR #1237, opened 2025-03-25
bpf_metadata: Add support for configurable SO_LINGER option

jrajahalme:
Add option bpf_metadata filter config 'original_source_so_linger_time'. If set to non-negative value, a SO_LINGER option is set on the upstream connections for the filter chain to the indicated value whenever an original source address is used.

Note: positive values make the worker thread serving the connection block during the close system call which can have a performance impact.


cilium/api/bpf_metadata.proto
```diff
@@ -55,4 +55,13 @@ message BpfMetadata {
   // l7lb_policy_name is the name of the L7LB policy that is enforced on the listener.
   // This is optional field.
   string l7lb_policy_name = 10;
+
+  // original_source_so_linger_time specifies the number of seconds to linger on socket close.
+  // Only used if use_original_source_address is also true, and the original source address
```
sayboras, line 60: From the agent side, we probably need to set this _always_ whenever use_original_source_address is true right ?

sayboras (approved):

cilium/api/bpf_metadata.proto
sayboras: ah just saw this one https://github.com/cilium/cilium/pull/38500

PR #1242, opened 2025-03-28
sni: Add logic to allow pattern prefix match

sayboras:
This commit is to support domain matching logic for SNI.

Examples of valid matches:

"*.example.com" matches:
  - "foo.example.com"
  - "bar.example.com"
  - "sub.example.com" 
but not:
  - "example.com" as wildcard requires exactly one label
  - "foo.bar.example.com" due to too many labels
  - "notexample.com" due to different domain

### DIFF
--- a/cilium/network_policy.h
+++ b/cilium/network_policy.h
@@ -344,5 +344,42 @@ protected:
 };
 using NetworkPolicyMapSharedPtr = std::shared_ptr<const NetworkPolicyMap>;
 
+struct SNIPattern {
+  std::string pattern;
+
+  explicit SNIPattern(const std::string& p) : pattern(absl::AsciiStrToLower(p)) {}
+
+  bool matches(const absl::string_view sni) const {
+    if (pattern.empty() || sni.empty()) {
+      return false;
+    }
+    auto const lower_sni = absl::AsciiStrToLower(sni);
+    // Perform lower case exact match if there is no wildcard prefix
+    if (!pattern.starts_with("*")) {
+      return pattern == lower_sni;
+    }
+
+    // Pattern is "**.<domain>"
+    if (pattern.starts_with("**.")) {
+      return lower_sni.ends_with(pattern.substr(2));
+    }
+
+    // Pattern is "*.<domain>"
+    if (pattern.starts_with("*.")) {
+      auto const sub_pattern = pattern.substr(1);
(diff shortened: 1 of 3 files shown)


cilium/network_policy.h
```diff
@@ -344,5 +344,42 @@ class NetworkPolicyMap : public Singleton::Instance,
...
+      return pattern == lower_sni;
+    }
+
+    // Pattern is "**.<domain>"
+    if (pattern.starts_with("**.")) {
+      return lower_sni.ends_with(pattern.substr(2));
+    }
+
+    // Pattern is "*.<domain>"
+    if (pattern.starts_with("*")) {
```
jrajahalme, line 368: IMO we should not allow patterns without the dot after the asterisk, so this should be `starts_with("*.")` instead.

tests/cilium_network_policy_test.cc
```diff
@@ -1726,5 +1726,59 @@ TEST_F(CiliumNetworkPolicyTest, TlsPolicyUpdate) {
...
+  EXPECT_TRUE(double_wildcard.matches("foo.bar.sub.example.com"));
+
+  // Test with wildcard label
+  SNIPattern wildcard_label("*example.com");
+  EXPECT_FALSE(wildcard_label.matches("foo.example.com"));
+  EXPECT_FALSE(wildcard_label.matches("bar.example.com"));
+  EXPECT_FALSE(wildcard_label.matches("FOO.EXAMPLE.COM"));
+  EXPECT_FALSE(wildcard_label.matches("example.com"));
+  EXPECT_FALSE(wildcard_label.matches("foo.bar.example.com"));
+  EXPECT_TRUE(wildcard_label.matches("fooexample.com"));
```
jrajahalme, line 1779: This should also be an `EXPECT_FALSE`

jrajahalme (changes requested):
We should remove the partial subdomain match, it is dangerous.

cilium/network_policy.h
sayboras: thanks and updated :heavy_check_mark:

tests/cilium_network_policy_test.cc
sayboras: Thanks, I still keep the whole negative test cases :+1:

jrajahalme (approved):
Good work, thank you :-)

PR #1257, opened 2025-04-04
build: Remove godeps target

sayboras:
This can be removed as upstream is updated with recent versions. Also, use the go tool directive to manage dependencies.


mhofstetter (approved):
:rocket:

PR #1266, opened 2025-04-09
ipcache: Make ipcache pin name configurable

jrajahalme:
Make ipcache pin file name configurable via bpf_metadata filter config.

Requires Cilium 1.18+ to explicitly set the new config option, older releases work on the (old) default name "cilium_ipcache".

This should work better with Cilium upgrades and downgrades. Version with this option needs to run with Cilium 1.18, but downgrade to prior Cilium versions should also work, as the IPcache is re-opened with the new (old) path name when Network Policy xDS stream is re-opened.

FIxes: #1190


sayboras (commented):
Do we still need https://github.com/cilium/proxy/pull/1255 ?

jrajahalme:
The fact that CI passed while using Cilium 1.18 must have been due to fallback to sending IP/ID mappings via protobuf if ipcache can not be opened.

sayboras (approved):

PR #1272, opened 2025-04-10
[v1.32] image: Bump ubuntu version to 24.04

sayboras:
Once Cilium 1.16 is the min supported version, we can move to ubuntu 24.04.


nickolaev (approved):
Looks good to me.

PR #1274, opened 2025-04-10
ipcache: reopen if name changes

jrajahalme:
On Cilium agent restart we have no control on the order of NetworkPolicy and Listener xDS stream restarts, so we could re-open the ipcache using the old path if the path is changing.


cilium/ipcache.cc
```diff
@@ -92,12 +92,19 @@ IPCache::IPCache(const std::string& path)
 
 void IPCache::SetPath(const std::string& path) {
   Thread::LockGuard guard(path_mutex_);
-  path_ = path;
+  if (path != path) {
```
sayboras, line 95: is it path_ != path ?

jrajahalme: good call!

jrajahalme: Failed test was at: https://github.com/cilium/proxy/actions/runs/14383419488/job/40334342351?pr=1274

jrajahalme:
fixed incorrect comparison.

cilium/ipcache.cc
sayboras: ah right, I didn't see any failure till now, let me take a look

sayboras (approved):

cilium/ipcache.cc
jrajahalme: Maybe these tests were never run with flow validation on? Could be simple as that?

sayboras: > Maybe these tests were never run with flow validation on? Could be simple as that?

True. The upstream cilium/cilium didn't have flow validation enabled for any workflow. We can disable it for the time being here as well.

PR #1285, opened 2025-04-22
policy: Add proxy_id matching

jrajahalme:
Add 'proxy_id' field to NetworkPolicy xDS 'PortNetworkPolicyRule'. If non-zero, the rule will be ignored when executing on a Listener for a non-matching proxy ID.


mhofstetter (approved):

jrajahalme:
rebased for merge conflicts

PR #1301, opened 2025-05-05
renovate: Correct bazelisk datasource

sayboras:


.github/renovate.json5
```diff
@@ -11,6 +11,7 @@
     'Dockerfile.builder',
     'go.mod',
     'go.sum',
+    'install_bazelisk.sh',
```
mhofstetter, line 14: -> `tools/install_bazelisk.sh`?

mhofstetter (changes requested):

.github/renovate.json5
sayboras: thanks, good catch :+1:

mhofstetter (approved):

PR #1302, opened 2025-05-05
envoy: Bump envoy version to 1.34.x

sayboras:
This is part of regular maintenance. Only merge once 1.18 is released, or 1.32 is EOL.

https://github.com/envoyproxy/envoy/releases/tag/v1.34.0


jrajahalme (approved):
Thanks Tam!

I ran `make tests` locally (on arm64) without any issues.

Maybe add a reference to the envoy toolchain config into `patches/0007-liburing-arm-build.patch` so that the flags are less likely to diverge? (even smaller nit: the new patch file is created in an earlier commit for no apparent reason).

patches/0007-liburing-arm-build.patch
```diff
@@ -0,0 +1,37 @@
...
++        "//bazel:linux_aarch64": {
++            "CFLAGS": "--target=aarch64-unknown-linux-gnu -fuse-ld=lld",
++            "CPPFLAGS": "--target=aarch64-unknown-linux-gnu -fuse-ld=lld",
++            "LDFLAGS": "-Wl,-S --target=aarch64-unknown-linux-gnu -fuse-ld=lld -Wl,-no-as-needed -Wl,-z,relro,-z,now -lm -l:libstdc++.a -lc",
++            "CC": "/usr/bin/clang-17",
++            "CXX": "/usr/bin/clang-17",
++            "LD": "/usr/bin/lld-17",
++        },
++        "//conditions:default": {},
++    }),
```
jrajahalme, line 31: Just realized that this was just recently removed in main. Any ideas why this would not be needed for Envoy 1.33, but then needed again in 1.34?

sayboras: I think it's related to the new feature introduced in upstream, will dig further once I have time.

> Added support for io_uring in the default socket interface.

PR #1336, opened 2025-05-13
ingress: Use ingress IPs as socket source address

sayboras:
The resolved backend IP could be the same client IP, which causes the
packet drop. This commit is to use the socket original source IP as
Ingress IP when use-original-source-address is set as true, just a note
that the policy enforcement is based on EndpointIP in socket mark instead.

Relates: https://github.com/cilium/cilium/pull/39530
Relates: https://github.com/cilium/cilium/issues/39531


sayboras:
closed in favour of https://github.com/cilium/proxy/pull/1336

PR #1337, opened 2025-05-14
conn: Skip original source address in hairpinning

sayboras:
Testing was done locally, in this PR, also in upstream https://github.com/cilium/proxy/pull/1337

### DIFF
--- a/cilium/bpf_metadata.cc
+++ b/cilium/bpf_metadata.cc
@@ -39,6 +39,7 @@
 #include "cilium/api/bpf_metadata.pb.h"
 #include "cilium/api/bpf_metadata.pb.validate.h" // IWYU pragma: keep
 #include "cilium/conntrack.h"
+#include "cilium/filter_state_cilium_destination.h"
 #include "cilium/filter_state_cilium_policy.h"
 #include "cilium/host_map.h"
 #include "cilium/ipcache.h"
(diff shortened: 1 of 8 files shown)


cilium/bpf_metadata.cc
```diff
@@ -552,13 +552,15 @@ Network::FilterStatus Instance::onAccept(Network::ListenerFilterCallbacks& cb) {
...
     cb.filterState().setData(
-        Cilium::CiliumPolicyFilterState::key(), socket_metadata->buildCiliumPolicyFilterState(),
+        Cilium::CiliumPolicyFilterState::key(), policy_fs,
         StreamInfo::FilterState::StateType::ReadOnly, StreamInfo::FilterState::LifeSpan::Connection,
         StreamInfo::StreamSharingMayImpactPooling::SharedWithUpstreamConnection);
 
     // Restoring original source address on the upstream socket
-    socket_options->push_back(socket_metadata->buildSourceAddressSocketOption(config_->so_linger_));
+    socket_options->push_back(
+        socket_metadata->buildSourceAddressSocketOption(config_->so_linger_, policy_fs));
```
mhofstetter, line 563: is it possible to avoid passing the policy filterstate to the socket option and instead looking it up once the socket option logic gets executed?

cilium/network_filter.cc
```diff
@@ -162,6 +162,8 @@ Network::FilterStatus Instance::onNewConnection() {
       ENVOY_CONN_LOG(warn, "cilium.network (egress): No destination address", conn);
       return false;
     }
+
+    policy_fs->dst_address = dst_address;
```
mhofstetter, line 166: Isn't it possible to extract the correct dst address when building the policy filterstate? 

i'm asking because previous PRs tried to create dedicated fitlerstates and socketoptions. I kind of fear that we're on the way of re-adding more and more properties/functionality to the existing "components" instead of adding new one.

alternatively we could create a new filterstate that is only responsible to transport this information.

cilium/bpf_metadata.cc
sayboras: iiuc, socket option is pretty low level, while filterstate is in the connection level, so I don't know how.

cilium/network_filter.cc
sayboras: having a dedicated filterstate is awesome point, I don't need to mess with readonly or mutable type for existing policy fs, thanks :+1:

cilium/bpf_metadata.cc
sayboras: the above was considered, however, imo using filter state is cleaner, especially after your comment on dedicated filter state.

jrajahalme: IMO this is now resolved with the dedicated filter state object.

PR #1339, opened 2025-05-15
cilium: change comment to test clang-tidy workflow

jrajahalme:


sayboras:
Seems like this is no longer required, feel free to re-open if you think otherwise.

PR #1355, opened 2025-05-21
ci: run cilium tests concurrently

viktor-kurchenko:
Not a big deal but it's a bit faster now:
- [Before](https://github.com/cilium/proxy/actions/runs/15165303106/job/42641420030): ~13 min 
- [After](https://github.com/cilium/proxy/actions/runs/15166520621/job/42645662595?pr=1355): ~6 min


sayboras (approved):
Thanks, it's awesome :heavy_check_mark:

PR #1371, opened 2025-05-29
bpf_metadata: Skip using original source address if destination is local host

jrajahalme:
Original source address need not and can not be used if the destination is a local pod, or a local host. Add the local host case to bpf_metadata filter, and add handling of both cases to socket_option_source_address, as at that time the chosen destination is known.

In principle, we should not need these checks in bpf_metadata any more, but tests/ currently depends on them.


jrajahalme:
resolved conflict, no longer using a `weak_ptr` for the policy resolver.

sayboras:
Closing this due to inactivity, feel free to re-open if it's still required.

PR #1372, opened 2025-05-29
Change policy resolver back to a shared pointer

jrajahalme:
Change policy resolver back to a shared pointer, and manage `NetworkPolicyMap` destruction explicitly by splitting the map to singleton and `NetworkPolicyMapImpl` and by posting the `NetworkPolicyMapImpl` to the main thread for deletion. This effectively reverts https://github.com/cilium/proxy/pull/1016.

`NetworkPolicyMapImpl` is managed via an unique_ptr, which required the removal of use of `shared_from_this`, which was possible due to simplified clean-up of the old map now that an atomic swap is used for the policy map update.

This change allows long lived connections on a removed listener to keep on functioning, including receiving policy updates via the singleton policy map.


jrajahalme:
cleaned up comments

PR #1385, opened 2025-06-04
l7lb: Pass original source address in option for east/west

jrajahalme:
Source address socket option uses 'hashKey' function to choose the
upstream connection pool. 'hashKey' has no access to the socket (which
has not been created yet), so it needs access to the original source
address which can be either IPv4 or IPv6 to be able to pass in the
original source address and port in the resulting hash.

This fixes the issue with east/west l7lb that caused upstream connection
pool being shared between all local endpoints with the same security
identity, while using the original source address. This manifested as
connection pool attempting to create more than one connection with the
same 5-tuple, which failed and resulted in '503' responses in HTTP.

Fixes: #65

### DIFF
--- a/cilium/bpf_metadata.cc
+++ b/cilium/bpf_metadata.cc
@@ -304,30 +304,19 @@ uint32_t Config::resolveSourceIdentity(const PolicyInstance& policy,
   return source_identity;
 }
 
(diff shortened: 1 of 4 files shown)


sayboras (approved):
oh it's tricky bug, nice.

PR #1393, opened 2025-06-11
cilium: Reduce locking in privileged service client

jrajahalme:
Change the calls to "privileged services" to only take the `calls_mutex_` for receiving responses.

Set up a `Waiter` for each caller so that the active receiver can pass responses to them. The active receiver stops receiving as soon as it gets its own response, and passes the receiver role to another "waiter".

Any changes to this code should be tested with a performance tool maximally loading all the available worker threads (e.g., `wrk`).

This partially addresses the performance regression introduced in #315.

On a local kind environment this improves `wrk -t50 -c100` request rate through a CEC-defined Envoy listener from 6221 req/sec to 13593 req/sec.

Fixes: #315

### DIFF
--- a/cilium/BUILD
+++ b/cilium/BUILD
@@ -359,6 +359,7 @@ envoy_cc_library(
     deps = [
         "//starter:privileged_service_protocol",
         "@envoy//envoy/api:os_sys_calls_interface",
+        "@envoy//source/common/common:thread_lib",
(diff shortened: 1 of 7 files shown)


jrajahalme:
applying some cleanups, converted to draft until stabile.

sayboras (approved):

PR #1395, opened 2025-06-11
bpf_metadata: Add config `use_nphds`

jrajahalme:
Add new proto config option `use_nphds`, which, if`true`, will use the
Network Policy Hosts xDS protocol to pull IP/ID mappings from Cilium
Agent when use of bpf maps is not configured (i.e., `bpf_root` is left empty).

This retains the current behavior of not initializing the Network Policy
map if `bpf_root` is empty, and allows enforcement of Cilium Network
Policies without access to bpf maps by specifying the new `use_nphds` as
'true' in that case.


mhofstetter (approved):
Thanks!

PR #1400, opened 2025-06-16
Policy refactor to fix logging, avoid unnecessary steps

jrajahalme:
Prepare network filter for future dual upstream/downstream change by only using the current upstream callback for L7 LB enforcement. Now that network level policy for non-L7 LB is performed directly in the downstream filter's onNewConnection() callback, the connection will be rejected on network policy deny before the HTTP level policy is enforced. Some tests needed to be updated to reflect this change by waiting for disconnect rather than HTTP '403' in these cases. This change also causes HTTP access logging to not happen to any traffic that is already dropped by the network level policy. Note that for L7 LB (i.e., Ingress, Gateway API) there is no change.

The case where this change could have production difference is when the network level policy rule enforces an SNI, as in these cases we may have emitted a drop type access log with HTTP header information instead of just the network level drop message.

Fix debug logging where a HTTP ALLOW was logged even when the enforcement was actually skipped in the downstream HTTP filter. This was a logging defect only, so there is no production impact due to this fix.

Finally, skip HTTP policy enforcement if the policy has no HTTP rules. This is possible since the network level policy is always enforced before the HTTP policy (per request), so the HTTP level enforcement only makes a difference when there are HTTP rules.

This last change relies on the current network level policy being enforced on each HTTP request. This needs to be revisited when the network level policy is only enforced on new upstream connections.

This change improves the max rate of requests passing through a CEC defined listener by ~15%.

### DIFF
--- a/cilium/accesslog.cc
+++ b/cilium/accesslog.cc
@@ -83,6 +83,8 @@ void AccessLog::Entry::initFromConnection(
     const std::string& policy_name, uint32_t proxy_id, bool ingress, uint32_t source_identity,
     const Network::Address::InstanceConstSharedPtr& source_address, uint32_t destination_identity,
     const Network::Address::InstanceConstSharedPtr& destination_address, TimeSource* time_source) {
+  request_logged_ = false;
+
   entry_.set_policy_name(policy_name);
   entry_.set_proxy_id(proxy_id);
   entry_.set_is_ingress(ingress);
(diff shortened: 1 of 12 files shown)


jrajahalme:
Rebased to not access log pod and Ingress policy allow separately for hairpinned Gateway traffic (which was introduced in the earlier version of this PR).

mhofstetter (approved):
Nice, thanks. Maybe @sayboras wants to take a second look

edit: and thanks for splitting this into individual commits :pray:

PR #1428, opened 2025-07-02
[v1.33] bpf_metadata: Skip original source address for same node L7 LB destinations

jrajahalme:
We already skip using the original source address/port for non-L7 LB destinations in the same node. This is needed to avoid 5-tuple collisions in cases where the destination happens to have an HTTP policy.

Do the same for east/west L7 LB (e.g., GAMMA) if the chosen destination is in the same node.

Same node destinations receive the source security identity via an skb_mark, so the original source address is not needed for policy enforcement at the destination.


sayboras:
The upstream PR https://github.com/cilium/proxy/pull/1437 is merged now, feel free to mark this as ready. We can bundle this update in coming patch release as well.

sayboras (approved):

PR #1437, opened 2025-07-04
bpf_metadata: Skip original source address for same node L7 LB destinations

jrajahalme:
We already skip using the original source address/port for non-L7 LB destinations in the same node. This is needed to avoid 5-tuple collisions in cases where the destination happens to have an HTTP policy.

Do the same for east/west L7 LB (e.g., GAMMA) if the chosen destination is in the same node.

Same node destinations receive the source security identity via an skb_mark, so the original source address is not needed for policy enforcement at the destination.

### DIFF
--- a/cilium/socket_option_source_address.cc
+++ b/cilium/socket_option_source_address.cc
@@ -18,6 +18,7 @@
 
 #include "absl/numeric/int128.h"
 #include "cilium/filter_state_cilium_destination.h"
+#include "cilium/filter_state_cilium_policy.h"
 
 namespace Envoy {
 namespace Cilium {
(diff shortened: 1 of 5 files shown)


cilium/socket_option_source_address.cc
```diff
@@ -77,15 +80,23 @@ bool SourceAddressSocketOption::setOption(
...
+                "Skipping restore of local address on socket: {} - source address is same as "
+                "destination address {}",
+                socket.ioHandle().fdDoNotUse(), dst_addr);
+      return true;
+    }
+    // Also skip using the original source address if destination is local.
+    // Local destinations have a policy configured on their address.
+    if (policy_fs_ && policy_fs_->exists(dst_addr)) {
+      ENVOY_LOG(trace, "Skipping restore of local address on socket: {} - destination is local {}",
+                socket.ioHandle().fdDoNotUse(), dst_addr);
```
sayboras, line 97: is there any chance the dest_addr is different with source_address i.e. 2 different pods on the same pod ?

jrajahalme: We have a policy entry keyed with each address of the (local) endpoint (e.g., IPv4 and IPv6). I don't think we currently can have two IPv4 addresses or two IPv6 addresses on any given endpoint, but if we do in the future, then we'd have to make sure each of them is reported in the NPDS update for the endpoint. This means that for every local endpoint IP there should be a policy configured, and we can use that fact to distinguish which endpoints are local and which are not, like here. Furthermore, we avoid using the original source address here in order to avoid 5-tuple collisions in the host networking namespace in case the destination pod happens to (also) have a HTTP policy (Envoy redirection). That can only happen if there is a policy for the destination endpoint.

sayboras (approved):

PR #1451, opened 2025-07-15
policy: Support rule precedence

jrajahalme:
Add new 'precedence' field.

Fix handling of deny rules when looking to TLS context and proxylib reference.

Storage of rules for any given port range is flattened to a single sorted vector, which makes dealing with deny rules and precedence field much more straightforward. This restructuring is visible in the policy debug formatting.

### DIFF
--- a/cilium/api/npds.proto
+++ b/cilium/api/npds.proto
@@ -129,6 +129,11 @@ message TLSContext {
 message PortNetworkPolicyRule {
   reserved 1; // used in Cilium versions upto 1.14
 
+  // Precedence level for this rule. Rules with **higher** numeric values take
+  // precedence, even over deny rules of lower precedence level.
+  // The lowest precedence (zero) is used when not specified.
+  uint32 precedence = 10;
+
   // Traffic on this port is denied for all `remote_policies` if true
   bool deny = 8;
 
--- a/cilium/api/npds.proto
+++ b/cilium/api/npds.proto
@@ -127,6 +127,8 @@ message TLSContext {
 // If all the predicates of a rule match a flow, the flow is matched by the
 // rule.
 message PortNetworkPolicyRule {
+  reserved 1; // used in Cilium versions upto 1.14
+
   // Traffic on this port is denied for all `remote_policies` if true
   bool deny = 8;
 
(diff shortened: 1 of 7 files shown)


cilium/network_policy.h
```diff
@@ -123,12 +122,9 @@ class PortPolicy : public Logger::Loggable<Logger::Id::config> {
   bool forRange(std::function<bool(const PortNetworkPolicyRules&, bool& denied)> allowed) const;
   bool forFirstRange(std::function<bool(const PortNetworkPolicyRules&)> f) const;
 
-  static bool hasHttpRules(const PolicyMap& map, const RulesList& wildcard_rules,
-                           const PolicyMap::const_iterator port_rules);
-
   const PolicyMap& map_;
-  const RulesList& wildcard_rules_;
-  const PolicyMap::const_iterator port_rules_; // iterator to 'map_'
+  const PortNetworkPolicyRules* wildcard_rules_;
```
nezdolik, line 126: why not use smart pointer here and below?

tests/cilium_network_policy_test.cc
```diff
@@ -1955,5 +1948,108 @@ TEST_F(CiliumNetworkPolicyTest, SNIPatternMatching) {
   EXPECT_FALSE(wildcard_label.matches(""));
 }
 
+TEST_F(CiliumNetworkPolicyTest, OrderedRules) {
```
nezdolik, line 1951: Maybe we need a test for multiple rules and one of the rules that has precedence not specified?

cilium/network_policy.h
jrajahalme: So that we don't accidentally keep (last) reference to the policy object in a worker thread. When worker threads request the policy, they get it by reference, which is never stored across the function calls, so there are never dangling references.

cilium/api/npds.proto
```diff
@@ -127,6 +127,12 @@ message TLSContext {
 // If all the predicates of a rule match a flow, the flow is matched by the
 // rule.
 message PortNetworkPolicyRule {
+  reserved 1; // used in Cilium versions upto 1.14
+
+  // Precedence level for this rule. Rules with **higher** numeric values take
+  // precedence, even over deny rules of lower precedence level.
+  uint32 precedence = 10;
```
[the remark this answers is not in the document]
jrajahalme: Will add:
```
// The lowest precedence (zero) is used when not specified.
```

cilium/network_policy.cc
```diff
@@ -653,24 +642,71 @@ using PortNetworkPolicyRuleConstSharedPtr = std::shared_ptr<const PortNetworkPol
 class PortNetworkPolicyRules : public Logger::Loggable<Logger::Id::config> {
```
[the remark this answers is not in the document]
jrajahalme: Would need to split up if we added more piece-meal unit tests, though.

nezdolik:
modulo the comments (bunch of which turned out to be no problem) lgtm

cilium/network_policy.h
jrajahalme: Added comment to state the above.

jrajahalme:
Merged on basis of Kateryna's review!

PR #1452, opened 2025-07-16
update github runners to oci gh arc runners

koksay:
CNCF has hosted ephemeral GitHub runners on Oracle that we want projects to use instead of the GitHub-hosted ones, which now incur a cost to use.

Please direct any questions to me, @jeefy, @krook, and @RobertKielty. You can also join `#cncf-ci-infra` channel on CNCF Slack.

> [!NOTE]
Oracle Cloud has a feature called OCPU, which comes with 2 VCPUs by default. The VM `oracle-vm-32cpu-128gb-x86-64` we provide has effectively 64 cores, so the number of CPUs is the same (but the memory is half of what you use at the moment)

### DIFF
--- a/.github/workflows/build-envoy-image-ci.yaml
+++ b/.github/workflows/build-envoy-image-ci.yaml
@@ -17,7 +17,7 @@ jobs:
   build-and-push-prs:
     timeout-minutes: 360
     name: Build and push multi-arch images
-    runs-on: ubuntu-latest-64-cores-256gb
+    runs-on: oracle-vm-32cpu-128gb-x86-64
     outputs:
       sha: ${{ steps.tag.outputs.sha }}
     steps:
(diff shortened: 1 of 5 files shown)


sayboras:
Thanks for your PR. 

The related workflow is with pull_request_target, so I have created the test PR as per below. It seems like the runner is never ready, can you help to check? Just a note that we have seen the same issue in the past.

PR: https://github.com/cilium/proxy/pull/1454
Sample run: https://github.com/cilium/proxy/actions/runs/16333999228/job/46142426038?pr=1454
Past issue: https://github.com/cilium/proxy/pull/1195

koksay:
@sayboras yeah, since a new vm is created for each job, it takes some time to start. 

Thanks for the merge 🖖

Issue #1168, opened 2025-02-11
CI: Run Cilium gateway api conformance test in PR


sayboras:
Currently, we are running cilium connectivity tests in PR, however, there are some more use cases from Gateway API.

Relates: https://github.com/cilium/proxy/pull/1166

PR #1458, opened 2025-07-18
gha: Add Cilium Gateway API tests

sayboras:
Relates: #1168

---

Testing is done with the temp commit (due to pull_request_target)

https://github.com/cilium/proxy/actions/runs/16363994416?pr=1458


mhofstetter (approved):
Seems like this PR introduces a lot of redundancy with the existing integration test where we run the connectivity tests (e.g. building cilium agent and operator).

The initial need for this was to build the Cilium Agent image with the envoy binary of the source of the PR. But in the meantime we could probably simplify this and use the Envoy daemonset mode in combination with the latest cilium agent & operator images and just use the cilium-proxy image of the PR for the cilium-proxy daemonset. 

But this shouldn't block this PR. thanks!

PR #1461, opened 2025-07-18
bpf_metadata: skip src address restore for N/S L7LB to dest ID world

mhofstetter:
Currently, source address restoration is done for destination IPs
with security identity WORLD (e.g. outside of the K8s cluster).

In case of N/S L7LB to cluster external IPs this breaks connectivity
because the node-local Cilium Ingress IP (that is set as source address
to integrate with CiliumNetworkPolicies) is "not known" to the outside world.

Therefore, source address restoration should be skipped in case of N/S
L7 loadbalancing for destination addresses with security identity WORLD.

Signed-off-by: Marco Hofstetter <marco.hofstetter@isovalent.com>


cilium/socket_option_source_address.cc
```diff
@@ -90,13 +91,26 @@ bool SourceAddressSocketOption::setOption(
...
     if (policy_fs_ && policy_fs_->exists(dst_addr)) {
       ENVOY_LOG(trace, "Skipping restore of local address on socket: {} - destination is local {}",
                 socket.ioHandle().fdDoNotUse(), dst_addr);
       return true;
     }
+
+    // Also skip using the original source address in case of N/S L7 loadbalancing if the
+    // destination identity is WORLD.
+    if (policy_fs_ && dest_fs_ && policy_fs_->is_l7lb_ && !original_source_address_ &&
+        policy_fs_->resolvePolicyId(dest_fs_->getDestinationAddress()->ip()) == Cilium::ID::WORLD) {
```
jrajahalme, line 106: This should likely be a bit wider condition, testing if the destination identity has the local scope bit set, as then this would cover WORLDIPv4 and WORLDIPv6, as well as any identities derived from toFQDNs or CIDR policies.

jrajahalme:
Another question to understand would be if it were possible to get the source IP masqueraded, or at least why is it not already masqueraded when exiting the node?

PR #1476, opened 2025-07-23
docker: Remove docker builder test

sayboras:
This is same as Docker.builder (except ubuntu version), let's keep one only to avoid multiple build steps.


mhofstetter (approved):
LGTM, thanks!

PR #1486, opened 2025-07-29
Clang bump to 18

jrajahalme:
Got some compilation errors for `make tests` locally that seemed to originate from a compiler/std library incompatibility. Best to use the same clang version upstream is using (https://github.com/envoyproxy/envoy/pull/38571), which is clang-18 for now.


Makefile
```diff
@@ -116,7 +116,7 @@ BUILD_DEP_HASHES: $(BUILD_DEP_FILES)
 clang.bazelrc: bazel/setup_clang.sh
 	$(call install_clang)
 	bazel/setup_clang.sh /usr/lib/llvm-17
-	echo "build --config=libc++" >> $@
+	echo "build --config=clang" >> $@
```
sayboras, line 119: nice, I noticed similar thing locally as well, the config is actually simplified in upstream

https://github.com/envoyproxy/envoy/releases/tag/v1.35.0
> Consolidated clang/gcc toolchains using --config=clang or --config=gcc.

sayboras (approved):
Thanks and lgtm. 

I have raised the issue for ubuntu bump later https://github.com/cilium/proxy/issues/1488.

PR #1487, opened 2025-07-29
tests: Wrap calls to logrus.Fatalf in a var

lomackie:
Maintaining the current behaviour in the test suite, this allows these fatal log calls to be overriden externally by fuzzers, as any exit call within a fuzzer is treated as an error by the CI.
See: https://github.com/cncf/cncf-fuzzing/pull/559 and https://github.com/cilium/cilium/pull/40728


sayboras (approved):
Ideally, we should not call fatal in tests, it's out of scope of this PR.

PR #1655, opened 2025-11-16
Add support for building release images with debug information

nickolaev:
This adds the bazel and make infrastructure to build release images with included symbols information.
It would be useful in production environments when debugging issues.

To leverage the build with CI, we move the core Release build functionality to `build-envoy-images-release-base.yaml` and then include it in the two new files:
* `build-envoy-images-release.yaml` - invoked on push in the branch
* `build-envoy-images-release-debug.yaml` - invoked on demand in the Actions UI

### DIFF
--- a/.github/workflows/build-envoy-images-release-base.yaml
+++ b/.github/workflows/build-envoy-images-release-base.yaml
@@ -1,0 +1,242 @@
+name: Refresh test & build cache & build latest
+on:
+  workflow_call:
+    inputs:
+      release_debug:
+        description: 'Build debug images'
+        required: false
+        type: boolean
+        default: false
+    secrets:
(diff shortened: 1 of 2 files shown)


.bazelrc
```diff
@@ -27,6 +27,13 @@ build --fission=dbg
 # upto-date
 build:release --define manual_stamp=manual_stamp
 
+# release_debug - an optimized release build containing the debug information
+build:release_debug -c opt
```
nezdolik: should we pass same flags as Envoy optimized build does? https://github.com/envoyproxy/envoy/blob/main/bazel/README.md?plain=1#L574

sayboras:
any reason to do it with 1.34 branch first ? Normally, we do the changes for main and then backport to older stable branch instead.

PR #1673, opened 2025-12-03
Add support for building release images with debug information 

nickolaev:
This adds the bazel and make infrastructure to build release images with included symbols information.
It would be useful in production environments when debugging issues.

To leverage the build with CI, we move the core Release build functionality to `build-envoy-images-release-base.yaml` and then include it in the two new files:
* `build-envoy-images-release.yaml` - invoked on push in the branch
* `build-envoy-images-release-debug.yaml` - invoked on demand in the Actions UI

### DIFF
--- a/.github/workflows/build-envoy-images-release-base.yaml
+++ b/.github/workflows/build-envoy-images-release-base.yaml
@@ -1,0 +1,242 @@
+name: Refresh test & build cache & build latest
+on:
+  workflow_call:
+    inputs:
+      release_debug:
+        description: 'Build debug images'
+        required: false
+        type: boolean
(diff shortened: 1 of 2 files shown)


mhofstetter:
@sayboras looks like `Run integration tests on amd64` fails with `Error when evaluating 'runs-on' for job 'tests'. .github/workflows/ci-tests.yaml (Line: 41, Col: 14): Unexpected value ''`

IMO this is related to this change https://github.com/cilium/proxy/pull/1578 - seems like the variable is missing in forks.

do we need to change the trigger from `pull_request` to `pull_request_target` ?

PR #1836, opened 2026-04-06
Delta npds

jrajahalme:
Implement support for delta NPDS and NPHDS updates.

NetworkPolicyMap switches to delta mode eagerly when there is evidence that the agent is capable (via BpfMetadata listener filter config), but we switch to SotW mode only when xDS stream transport had failed to connect or closes. This should work for Cilium Agent upgrades and downgrades, as the agent
expresses the desired mode, and upgraded agents listen for both SotW NPDS and Delta NPDS.

Network policy map maintains a stream generation number for new stream detection purposes. This is implemented using a new stream events callback added to upstream Envoy gRPC Mux classes via a new patch.

### DIFF
--- a/cilium/api/bpf_metadata.proto
+++ b/cilium/api/bpf_metadata.proto
@@ -86,4 +86,10 @@ message BpfMetadata {
 
   // Configuration for the source of NPDS updates. Currently this field is not supported.
   envoy.config.core.v3.ConfigSource npds_config = 16;
+
+  // Use delta NPDS rather than the state-of-the-world protocol.
+  // Even with delta NPDS, each new stream starts with a full dump.
+  // Only wildcard subscriptions are supported.
+  // All listeners on the node must agree on this setting.
+  bool use_delta_npds = 17;
 }
--- a/cilium/grpc_subscription.cc
+++ b/cilium/grpc_subscription.cc
@@ -2,16 +2,19 @@
 
 #include <fmt/format.h>
 
-#include <chrono>
+#include <cstdint>
+#include <functional>
 #include <memory>
 #include <string>
 #include <utility>
 #include <vector>
 
 #include "envoy/annotations/resource.pb.h"
+#include "envoy/common/callback.h"
 #include "envoy/common/exception.h"
 #include "envoy/config/core/v3/config_source.pb.h"
 #include "envoy/config/custom_config_validators.h"
+#include "envoy/config/grpc_mux.h"
 #include "envoy/config/subscription.h"
 #include "envoy/config/subscription_factory.h"
 #include "envoy/grpc/async_client.h"
@@ -21,13 +24,17 @@
 
 #include "source/common/common/assert.h"
 #include "source/common/common/backoff_strategy.h"
+#include "source/common/common/logger.h"
 #include "source/common/config/utility.h"
 #include "source/common/grpc/common.h"
 #include "source/common/protobuf/protobuf.h" // IWYU pragma: keep
 #include "source/extensions/config_subscription/grpc/grpc_mux_context.h"
+#include "source/extensions/config_subscription/grpc/grpc_mux_impl.h"
 #include "source/extensions/config_subscription/grpc/grpc_subscription_impl.h"
+#include "source/extensions/config_subscription/grpc/new_grpc_mux_impl.h"
 
 #include "absl/container/flat_hash_map.h"
+#include "absl/container/flat_hash_set.h"
 #include "absl/status/statusor.h"
 #include "absl/strings/match.h"
 #include "absl/strings/string_view.h"
@@ -102,7 +109,14 @@ public:
                          const Protobuf::RepeatedPtrField<std::string>&) override {}
 };
 
-} // namespace
+bool configSourceUsesDeltaXds(const envoy::config::core::v3::ConfigSource& config_source) {
+  if (!config_source.has_api_config_source()) {
+    return false;
+  }
+  const auto& api_type = config_source.api_config_source().api_type();
+  return api_type == envoy::config::core::v3::ApiConfigSource::DELTA_GRPC ||
+         api_type == envoy::config::core::v3::ApiConfigSource::AGGREGATED_DELTA_GRPC;
+}
 
 const Protobuf::MethodDescriptor& deltaGrpcMethod(absl::string_view type_url) {
   const auto it = typeUrlToServiceMap().find(static_cast<std::string>(type_url));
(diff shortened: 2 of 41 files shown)


cilium/api/bpf_metadata.proto
```diff
@@ -82,4 +82,10 @@ message BpfMetadata {
 
   // Cache is garbage collected at interval 10 times the ttl (default 30 ms).
   google.protobuf.Duration cache_gc_interval = 15;
+
+  // Use delta NPDS rather than the state-of-the-world protocol.
+  // Even with delta NPDS, each new stream starts with a full dump.
+  // Only wildcard subscriptions are supported.
+  // All listeners on the node must agree on this setting.
+  bool use_delta_npds = 16;
```
nezdolik: i think cleaner & more canonical way (which would allow plugging in multiple config source types) would be to use https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/config_source.proto, e.g.:
```
npds_config:
  api_config_source:
    api_type: GRPC
    ....
```
That would allow to cleanly introduce support for ads as well:
```
npds_config:
  ads: {}
```

jrajahalme:
Changed to clear the resource map on new streams before parsing policies. This forces re-parsing policies that have unchanged to use the current subscription's ConfigSource for Secret watchers.

cilium/host_map.cc
```diff
@@ -159,6 +173,29 @@ struct ThreadLocalHostMapInitializer : public PolicyHostMap::ThreadLocalHostMap
...
+      }
+      if (map.empty()) {
+        vec_it = maps.erase(vec_it);
+      } else {
+        ++vec_it;
+      }
+    }
+  }
+
+  void remove(const absl::flat_hash_set<uint64_t> removed_nids) {
```
nezdolik: nit: add & to the function argument declaration

cilium/host_map.cc
```diff
@@ -197,23 +234,156 @@ PolicyHostMap::PolicyHostMap(Server::Configuration::CommonFactoryContext& contex
...
-                              context.api().randomGenerator(), *scope_, *this,
-                              std::make_shared<Cilium::PolicyHostDecoder>());
+  context_ = &context;
+  desired_config_source_ = npds_config;
+  subscribe();
+}
+
+void PolicyHostMap::setConfigSource(
+    const envoy::config::core::v3::ConfigSource& config_source) const {
+  auto* self = const_cast<PolicyHostMap*>(this);
```
nezdolik: hm, why is this const cast needed? since the method modifies the underlying object, better remove `const` qualifier from declaration?

cilium/host_map.h
```diff
@@ -229,22 +231,27 @@ class PolicyHostMap : public Singleton::Instance,
...
+  void onSubscriptionStreamEvent(uint64_t subscription_id, Config::GrpcMuxStreamEvent event);
+  void maybeRecreateSubscriptionInDesiredMode(bool transport_closed);
+
   ThreadLocal::SlotPtr tls_;
   std::string name_;
   Stats::ScopeSharedPtr scope_;
   Stats::ScopeSharedPtr stats_scope_;
   std::unique_ptr<Envoy::Config::Subscription> subscription_;
+  Server::Configuration::CommonFactoryContext* context_{nullptr};
+  envoy::config::core::v3::ConfigSource desired_config_source_;
```
nezdolik: nit: add explanatory comment why two config_sources are needed?

nezdolik: also is it possible to make them const?

nezdolik (commented):
did one quick pass, will need to do one more round.

cilium/host_map.cc
jrajahalme: Good point, don't want to be copying it :-) Fixed.

cilium/host_map.h
jrajahalme: The desired_config_source_ may be set to a pessimistic value via explicit BpfMetadata config in CiliumEnvoyConfig CRD, and we should not change to a "worse" (e.g., SotW) ConfigSource if "better" (e.g., Delta) is already up-and-running. Added a comment.

Can not be const as they are set via the setConfigSource().

cilium/host_map.cc
jrajahalme: ugly hack, fixed. Was quite a bit of work...

jrajahalme:
Fixed BpfMetadata `policy_type` from `bool` to `PolicyType` enum.

jrajahalme:
Dropped the NPRDS (aka NPDSv2) from this PR, as it needs more work to properly support named ports. So this PR now adds Delta xDS support for NPDS and NPHDS.

jrajahalme:
Combined middle commits to eliminate unnecessary churn.

jrajahalme:
Fixed source format.

cilium/grpc_subscription.cc
```diff
@@ -118,55 +135,201 @@ const Protobuf::MethodDescriptor& sotwGrpcMethod(absl::string_view type_url) {
...
+        false);
+    THROW_IF_NOT_OK_REF(factory_or_error.status());
+
+    absl::StatusOr<Config::RateLimitSettings> rate_limit_settings_or_error =
+        Config::Utility::parseRateLimitSettings(api_config_source);
+    THROW_IF_NOT_OK_REF(rate_limit_settings_or_error.status());
+
+    const auto& api_type = api_config_source.api_type();
+    bool use_delta = api_type == envoy::config::core::v3::ApiConfigSource::DELTA_GRPC ||
+                     api_type == envoy::config::core::v3::ApiConfigSource::AGGREGATED_DELTA_GRPC;
```
nezdolik, line 166: since we have not fully tested yet if AGGREGATED_DELTA_GRPC works, we should better only focus on DELTA_GRPC? and reject a config with `AGGREGATED_DELTA_GRPC` ?

cilium/grpc_subscription.h
```diff
@@ -1,51 +1,109 @@
...
 
+#include "absl/status/status.h"
 #include "absl/strings/string_view.h"
 
 namespace Envoy {
 namespace Cilium {
 
-// GrpcMux wrapper to get access to control plane identifier
-class GrpcMuxImpl : public Config::GrpcMuxImpl {
+bool configSourceUsesDeltaXds(const envoy::config::core::v3::ConfigSource& config_source);
```
nezdolik: looks like there is no need to export this, could be part of anon namespace in impl file

cilium/grpc_subscription.h
```diff
@@ -1,51 +1,109 @@
...
-  ~GrpcMuxImpl() override = default;
+  Stats::Scope& scope() { return *scope_; }
 
-  void onStreamEstablished() override {
-    new_stream_ = true;
-    Config::GrpcMuxImpl::onStreamEstablished();
-  }
+  bool usesDeltaXds() const { return configSourceUsesDeltaXds(config_source_); }
+  bool connected() const { return connected_; }
+  uint64_t streamGeneration() const { return stream_generation_; }
```
nezdolik, line 62: another approach (rather than declaring bunch of protected utility accessor methods here in header file) for being able to access object state in tests is to declare a friend class as class member, e.g. https://github.com/envoyproxy/envoy/blob/main/test/extensions/filters/http/geoip/config_test.cc#L23 and https://github.com/envoyproxy/envoy/blob/c58a365828c1d4c0086b5355fbec8f3e6ff08d27/source/extensions/filters/http/geoip/geoip_filter.h#L69. Makes it easier to read header file.

cilium/grpc_subscription.h
```diff
@@ -1,51 +1,109 @@
...
 
-std::unique_ptr<Config::Subscription>
-subscribe(const absl::string_view type_url,
-          const envoy::config::core::v3::ConfigSource& config_source,
-          Server::Configuration::CommonFactoryContext& context, Stats::Scope& scope,
-          Config::SubscriptionCallbacks& callbacks,
-          Config::OpaqueResourceDecoderSharedPtr resource_decoder,
-          std::chrono::milliseconds init_fetch_timeout = std::chrono::milliseconds(0));
+  const std::string type_url_;
+  envoy::config::core::v3::ConfigSource desired_config_source_;
```
nezdolik, line 76: think i asked this before and maybe you responded in comment, but would be worth adding comment explaining what `desired_config_source_` is being used for.

cilium/grpc_subscription.cc
```diff
@@ -118,55 +135,201 @@ const Protobuf::MethodDescriptor& sotwGrpcMethod(absl::string_view type_url) {
...
+    const Protobuf::RepeatedPtrField<std::string>& removed_resources,
+    const std::string& system_version_info) {
+  connected_ = true;
+  return onDeltaConfigUpdate(added_resources, removed_resources, system_version_info);
+}
+
+void ManagedGrpcSubscription::onConfigUpdateFailed(Config::ConfigUpdateFailureReason reason,
+                                                   const EnvoyException* e) {
+  if (e != nullptr) {
+    ENVOY_LOG(debug,
```
nezdolik: should this be logged at debug level? warn or error feels more appropriate, otherwise could be missed

cilium/host_map.cc
```diff
@@ -234,7 +231,7 @@ PolicyHostMap::onConfigUpdate(const std::vector<Envoy::Config::DecodedResourceRe
   // dangerous as it may happen that there is a circular reference
   // from 'this' to itself via the lambda capture, leading to 'this'
   // never being released. It should happen in this case, though.
-  std::shared_ptr<PolicyHostMap> shared_this = shared_from_this();
+  std::shared_ptr<ManagedGrpcSubscription> shared_this = shared_from_this();
```
nezdolik: did not notice consistent approach to type naming in our codebase, but we could declare in header file smth like:
```
using PolicyHostMapSharedPtr = std::shared_ptr<PolicyHostMap>;
...
```

cilium/network_policy.cc
```diff
@@ -1765,7 +1754,7 @@ void NetworkPolicyMapImpl::reopenIpcache() {
   // Cilium agent re-creates IP cache on restart, and the first accepted update on
   // the new stream must reopen it before workers enforce refreshed identities.
   IpCacheSharedPtr ipcache = IpCache::getIpCache(context_);
-  if (ipcache != nullptr) {
+  if (ipcache) {
```
nezdolik, line 1994: nit: there also does not seem to be consistency for nullptr type of check in codebase, but could be better to leave to previous style? (seen it elsewhere)

tests/bpf_metadata_integration_test.cc
```diff
@@ -149,96 +185,182 @@ class BpfMetadataIntegrationTest : public BaseIntegrationTest,
...
+  void createAdsStream() {
+    createXdsConnection();
+    createXdsStream(ads_stream_);
   }
 
-  void sendCdsResponse(const std::string& version) {
+  void createSotWStreams(const std::string& response_version) {
+    createXdsConnection();
+
+    for (int i = 0; i < 4; i++) {
```
nezdolik, line 314: nit: please add explanation why magic number 4 (i < 4).

cilium/grpc_subscription.h
```diff
@@ -1,51 +1,109 @@
...
+
+  ManagedGrpcSubscription(absl::string_view type_url, DecoderFactory decoder_factory,
+                          const envoy::config::core::v3::ConfigSource& config_source,
+                          Server::Configuration::CommonFactoryContext& context,
+                          Stats::ScopeSharedPtr scope, bool subscribe = true);
+  ~ManagedGrpcSubscription() override = default;
+
+  const envoy::config::core::v3::ConfigSource& getConfigSource() const { return config_source_; }
+
+  // must be configured before subscibe()
```
nezdolik: :s/subscibe/subscribe

cilium/grpc_subscription.cc
```diff
@@ -118,55 +135,201 @@ const Protobuf::MethodDescriptor& sotwGrpcMethod(absl::string_view type_url) {
...
+    }
+  }
+  subscribe();
+  start();
+}
+
+absl::Status
+ManagedGrpcSubscription::onConfigUpdate(const std::vector<Config::DecodedResourceRef>& resources,
+                                        const std::string& version_info) {
+  connected_ = true;
```
nezdolik: seems to be set to true in both here and `ManagedGrpcSubscription::onStreamEvent` ? Should be enough to have it in `ManagedGrpcSubscription::onStreamEvent(Established)`?

cilium/host_map.cc
```diff
@@ -202,10 +225,62 @@ absl::Status PolicyHostMap::onDeltaConfigUpdate(
...
+      std::make_shared<ThreadLocalHostMapInitializer>(is_new_stream ? nullptr : getHostMap());
+
+  absl::flat_hash_set<uint64_t> to_remove;
+  to_remove.reserve(added_resources.size() + removed_resources.size());
+
+  for (const auto& name : removed_resources) {
+    uint64_t nid = 0;
+    auto [ptr, ec] = std::from_chars(name.data(), name.data() + name.size(), nid);
+    if (ec != std::errc{} || ptr != name.data() + name.size()) {
+      throw EnvoyException(fmt::format("Invalid removed resource name '{}'", name));
```
nezdolik, line 247: does this run on main or worker thread?

cilium/grpc_subscription.cc
jrajahalme: Upgrade to an Delta ADS version will break if we reject it, turning it to SotW ADS instead.

cilium/grpc_subscription.h
jrajahalme: done.

cilium/grpc_subscription.h
jrajahalme: Done, that was ugly, much better now :-)

cilium/grpc_subscription.h
jrajahalme: Added this comment:
```
  // desired_config_source_ is used to store config updates from BpfMetadata filter.
  // Listeners can ask for different configs, so we may keep the existing config_source_
  // if it remains functional even if some BpfMetadata filters ask for something different.
```
This would currently happen with CiliumEnvoyConfigs that specify the BpfMetadata listener filter explicitly and leave out the `cilium_config_source` field. BpfMetadata filter then defaults that to `GRPC`, while we may already be up and running with a functional ADS or DELTA_GRPC subscription.

We should probably override the `cilium_config_source` in any such CEC/CCEC resource in the Cilium Agent, though.

cilium/grpc_subscription.cc
jrajahalme: Changed log to `warn`.

cilium/host_map.cc
jrajahalme: Added this. Many such cases still remain, but not the concern of this PR.

cilium/network_policy.cc
jrajahalme: I tend to see `if (ipcache)` as more readable...

tests/bpf_metadata_integration_test.cc
jrajahalme: it is just big enough to cover the streams used in the test. better than looping for ever if the desired condition never appears. Added a comment :-)

cilium/grpc_subscription.h
jrajahalme: Fixed, thanks!

cilium/grpc_subscription.cc
jrajahalme: Now that we call `stream_event_callback(Config::GrpcMuxStreamEvent::Established)` if the mux is connected before creating the subscription it may be enough. At some point I needed to use the `onConfigUpdate` calls getting through as evidence for a connected state. Investigating...

cilium/host_map.cc
jrajahalme: xDS config updates always execute in the main (or test) thread.

cilium/grpc_subscription.h
jrajahalme: `factory_for_test_` remains, I don't see a clean way to get rid of it without building in support for file based subscriptions that we do not need in production code?

jrajahalme: Other than maybe converting tests that currently use file based subscription to use the gRPC API to provide the test config instead? IMO that is off the scope of this PR, though.

jrajahalme:
@nezdolik Thanks for the review, ready for a 2nd look :-)

jrajahalme:
fixed format

nezdolik:
looks like more format fixes are needed

jrajahalme:
> looks like more format fixes are needed

No, it's just the tidy dying on the tiny runner, see: https://github.com/cilium/proxy/pull/1936

I'll rebase to see it that fixed it!

jrajahalme:
rebased to pick up #1936

nezdolik (approved):

PR #1838, opened 2026-04-09
[bpf_metadata] Make use of npds_config if configured

nezdolik:


cilium/bpf_metadata.cc
```diff
@@ -234,9 +235,10 @@ Config::Config(const ::cilium::BpfMetadata& config,
         fmt::format("cilium.bpf_metadata: ipv6_source_address is not an IPv6 address: {}",
                     config.ipv6_source_address()));
   }
-
+  const absl::optional<envoy::config::core::v3::ApiConfigSource> npds_config =
+      config.has_npds_config() ? absl::make_optional(config.npds_config()) : absl::nullopt;
```
jrajahalme: IMO we could lift up defaulting to the cilium_xds_api_config here and avoid the optional wrapper and the related boilerplate?

nezdolik:
converting to draft until i finish integration test

cilium/bpf_metadata.cc
nezdolik: to my preference is better to keep default xds config source details within [grpc_subscription.h](https://github.com/cilium/proxy/blob/main/cilium/grpc_subscription.cc#L150) file and not to pull them up into bpf_metadata file. Even if we default in bpf_metadata  config, the config source would still need to be passed down to subscription level.

nezdolik:
> converting to draft until i finish integration test

changed my mind due to the amount of effort. Will investigate this separately (how/if we can have a test that uses real bpf_metadata.proto instead of test proto)

nezdolik:
need to fix tests, since i removed some utility methods

cilium/secret_watcher.cc
```diff
@@ -32,15 +32,26 @@ namespace Cilium {
 namespace {
 
 // SDS config used in production
-envoy::config::core::v3::ConfigSource getCiliumSDSConfig(const std::string&) {
-  /* returned config_source has initial_fetch_timeout of 50 milliseconds. */
-  return Cilium::cilium_xds_api_config;
+envoy::config::core::v3::ConfigSource
+getCiliumSDSConfig(const std::string&,
+                   absl::optional<const envoy::config::core::v3::ConfigSource> config_source_opt) {
```
nezdolik, line 37: looks like this does not need to be optional

cilium/bpf_metadata.cc
nezdolik: reworked and moved to bpf_metadata level

nezdolik:
Closing in favor of https://github.com/cilium/proxy/pull/1861 (developed in upstrea, branch and not my own fork) for sake of successful ci runs

PR #1861, opened 2026-04-20
[bpf_metadata] Make use of npds_config if configured

nezdolik:
When `npds_config` config field is set in bpf_metadata,  override the existing default cilium_xds_api_config_source and use it to fetch network policies, network policy hosts and secrets. This is prerequisite for supporting ads in cilium proxy.


nezdolik:
@jrajahalme ptal

jrajahalme (approved):

PR #1869, opened 2026-04-23
make: Pull up-to-date clang 18.1.8

jrajahalme:
Sync "clang.bazelrc" make target with the Dockerfile.builder and pull the latest clang 18.1.8. Ubuntu 24.04 repositories have only 18.1.3, which causes Envoy crashes like this:

GWP-ASan Mismatched-size-delete of 24 bytes (expected 25 bytes)

This is due to incompatibility issue with std::basic_string implementation in libc++ in clang 18.1.3, where it can drop the lowest bit in the allocation length, causing delete to be called with length 24, when the length required of allocation was 25. This is fixes in later clang libc++ versoins.

Upstream Envoy is pinned on Clang 18.1.8, and Dockerfile.builder already pulls the latest in version 18, which is 18.1.8.

Fixes: #1820

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@
 
 include Makefile.defs
 
+MIN_CLANG_VERSION := 18.1.8
 COMPILER_DEP := clang.bazelrc
 
 ENVOY_BINS = cilium-envoy bazel-bin/cilium-envoy cilium-envoy-starter bazel-bin/cilium-envoy-starter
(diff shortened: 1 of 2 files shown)


nezdolik:
Could also be the rootcause for https://github.com/cilium/cilium/issues/45448 ?

nezdolik (approved):

jrajahalme:
> Could also be the rootcause for [cilium/cilium#45448](https://github.com/cilium/cilium/issues/45448) ?

Not likely, since the release builds have already been built with clang 18.1.8. Besides in the linked issue the crash is sooner than the allocation/delete size mismatch detection.

PR #1871, opened 2026-04-23
tls_wrapper: Initialize socket earlier for downstream/server connections

jrajahalme:
Downstream listeners build the filter chain after transport socket callbacks are installed, but our TLS wrapper instantiated the SSL socket only later, in `onConnected()` callback. This caused `HttpConnectionManager` to miscount SSL connection stats and underflow in the count on delete. This underflow is guarded by an `ASSERT` that is not enforced on production builds (`NDEBUG` defined). This is why this never tripped in production, but the SSL active connection stats would have been underflowing to large positive values prior to this fix.

Create the wrapped socket eagerly so `connection().ssl()` is already populated when network filters such as HCM initialize their SSL-specific connection stats.

### DIFF
--- a/cilium/tls_wrapper.cc
+++ b/cilium/tls_wrapper.cc
@@ -49,6 +49,17 @@ public:
   // Network::TransportSocket
   void setTransportSocketCallbacks(Network::TransportSocketCallbacks& callbacks) override {
     callbacks_ = &callbacks;
+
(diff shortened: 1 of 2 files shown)


nezdolik (approved):
no tests needed?

jrajahalme:
> no tests needed?

Added test, thanks for the nudge :-)

PR #1877, opened 2026-04-24
policy: More robust removeInitManager

jrajahalme:
Replace the nullptr init manager with a dedicated `parked_init_manager_`, and inspect it for unintended use so that we can log warnings as needed. This compiles without relying on undefined behavior, and still provides a signal if targets are added after initialization has completed.

### DIFF
--- a/cilium/network_policy.cc
+++ b/cilium/network_policy.cc
@@ -15,6 +15,7 @@
 #include <utility>
 #include <vector>
 
+#include "envoy/admin/v3/init_dump.pb.h"
 #include "envoy/common/exception.h"
 #include "envoy/common/matchers.h"
 #include "envoy/common/optref.h"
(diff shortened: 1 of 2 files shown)


cilium/network_policy.cc
```diff
@@ -1921,15 +1924,39 @@ bool NetworkPolicyMapImpl::isNewStream() {
 
 // removeInitManager must be called at the end of each policy update
 void NetworkPolicyMapImpl::removeInitManager() {
-  // Remove the local init manager from the transport factory context
-#ifdef __clang__
```
nezdolik, line 1925: happy to see this gone :)

cilium/network_policy.cc
```diff
@@ -1921,15 +1924,39 @@ bool NetworkPolicyMapImpl::isNewStream() {
...
-#endif
-  transport_factory_context_->setInitManager(*static_cast<Init::Manager*>(nullptr));
-#ifdef __clang__
-#pragma clang diagnostic pop
-#endif
+  RELEASE_ASSERT(parked_init_manager_ != nullptr, "parked init manager must exist");
+
+  envoy::admin::v3::UnreadyTargetsDumps parked_dumps;
+  parked_init_manager_->dumpUnreadyTargets(parked_dumps);
+  RELEASE_ASSERT(parked_dumps.unready_targets_dumps_size() == 1,
```
nezdolik: should this be a fatal check? or we could use `ENVOY_BUG` instead?

nezdolik (commented):
may be worth adding a test?

cilium/network_policy.cc
jrajahalme: On a second thought this should be a warning log instead. Not good to crash Envoy for this in production, if there are no other adverse effects. Warning logs should still surface in CI runs.

jrajahalme: I'll take that back, this would indeed be a good spot for ENVOY_BUG. We need to know ASAP if the upstream behavior changes so that we can adapt accordingly.

jrajahalme: Generalized this to work with any number of dumps, so no need for RELEASE_ASSERT or ENVOY_BUG here any more.

nezdolik (approved):

PR #1878, opened 2026-04-24
Tests: Reduce flakiness

jrajahalme:
This PR addresses possible causes for test flakes in the CI and local test runs. These tend to be in the area of test fixture teardown causing issues or blending with next test setup in unintended ways. Please review by commit.

As some of the flakiness was due to production code deficiencies, it would be nice to get to a state where we have low enough flake probability to let individual flakes fail the test runs in CI.

### DIFF
--- a/tests/cilium_tls_tcp_integration_test.cc
+++ b/tests/cilium_tls_tcp_integration_test.cc
@@ -4,30 +4,25 @@
 #include <gmock/gmock-spec-builders.h>
 #include <gtest/gtest-param-test.h>
 #include <gtest/gtest.h>
-#include <unistd.h>
 
 #include <chrono>
 #include <cstdint>
 #include <functional>
 #include <memory>
 #include <string>
-#include <utility>
 #include <vector>
 
 #include "envoy/buffer/buffer.h"
-#include "envoy/common/exception.h"
 #include "envoy/event/dispatcher.h"
-#include "envoy/extensions/transport_sockets/tls/v3/tls.pb.h"
 #include "envoy/http/codec.h" // IWYU pragma: keep
 #include "envoy/network/address.h"
 #include "envoy/network/connection.h"
 #include "envoy/network/transport_socket.h"
+#include "envoy/ssl/connection.h"
 
 #include "source/common/buffer/buffer_impl.h"
+#include "source/common/buffer/watermark_buffer.h"
 #include "source/common/common/assert.h"
-#include "source/common/stats/isolated_store_impl.h"
-#include "source/common/tls/server_context_config_impl.h"
-#include "source/common/tls/server_ssl_socket.h"
 
 #include "test/integration/fake_upstream.h"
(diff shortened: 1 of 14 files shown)


tests/uds_server.cc
```diff
@@ -37,38 +36,52 @@ UDSServer::UDSServer(const std::string& path, std::function<void(const std::stri
...
     return;
   }
+}
 
-  ENVOY_LOG(trace, "Starting unix domain socket server thread fd: {}", fd_.load());
+UDSServer::~UDSServer() { shutdownServerThread(); }
 
+void UDSServer::startServerThread() {
+  if (fd_ < 0 || thread_ != nullptr) {
+    return;
```
nezdolik, line 60: nit: could add a log warning here that thread was not actually started

tests/cilium_tls_tcp_integration_test.cc
```diff
@@ -179,6 +176,10 @@ class CiliumTLSIntegrationTest : public CiliumTcpIntegrationTest {
           ON_CALL(*client_write_buffer_, drain(_))
               .WillByDefault(Invoke(client_write_buffer_, &MockWatermarkBuffer::trackDrains));
           return client_write_buffer_;
+        }))
+        .WillRepeatedly(Invoke([](std::function<void()> below_low, std::function<void()> above_high,
```
nezdolik, line 180: why does the buffer need to be allocated twice?

nezdolik (approved):
lgtm with 2 nits

tests/cilium_tls_tcp_integration_test.cc
jrajahalme: That `.WillRepeatedly` is there because `mock_buffer_factory_` is dispatcher-wide, not just for the one SSL client write buffer we care about.

The test wants the first buffer created during `createClientConnection()` to be a `MockWatermarkBuffer`, so it can track actual drained bytes here:
```
    while (client_write_buffer_->bytesDrained() != data_to_send_upstream.size()) {
```
But the connection/dispatcher/TLS stack may ask the same factory for additional buffers later. Those extra buffers should be ordinary `Buffer::WatermarkBuffers`. Without the repeated fallback, the `EXPECT_CALL(...).WillOnce(...)` is exhausted after the tracked buffer, and later buffer allocations either trip gMock “actions ran out” behavior or risk returning the wrong/default thing.

So the pattern means: first buffer is the instrumented one; every subsequent buffer behaves normally. It mirrors [Envoy’s IntegrationTcpClient setup](https://github.com/envoyproxy/envoy/blob/main/test/integration/integration_tcp_client.cc#L52-L54).

The need for this came up tracing a flake that happened under repeated test runs.

tests/uds_server.cc
jrajahalme: This was defensive return in case opening socket errored out or the thread was already started. Since this is only test code I'll let this be without a warning log for now.

PR #1887, opened 2026-05-06
tests: Fix cilium_websocket_policy_integration_test flakes

jrajahalme:
cilium_websocket_policy_integration_test still flaked when built as non-release and with multiple runs in parallel. Fixes:

- let the upstream test framework set the cluster port with the fake backend port
- Replace static listener ports with dynamically allocated ones
- sample the ping count before connecting in CiliumWebSocketUpstreamWritesFirst, wait for an uptick after
- remove ping waits from tests that already have proof of roundtrip traffic (hello/world)


sayboras (approved):
Running test locally few times, all runs are good 💯

Issue #1763, opened 2026-02-27
Cleanup proxylib related code once cilium 1.18 EOL


sayboras:
Proxylib support is removed in cilium/cilium upstream as part of 1.19 dev cycle. This issue is to track the work of removing proxylib related code in this repo once cilium 1.18 is EOL.

Relates: https://github.com/cilium/cilium/pull/43557

PR #1889, opened 2026-05-06
fix: avoid duplicate memcached denial queueing

immanuwell:
Fixes a small binary memcached parser bug.

When a denied request had to wait behind an earlier allowed response, the parser queued that denied response twice. First one got injected, but the stale duplicate stayed at the front of the queue. After that, the next delayed denial could get stuck behind it. tiny thing, but yeah, not great.

How to reproduce the bug on the old code:

1. send an allowed binary memcached request
2. send a denied binary memcached request
3. receive the allowed response, this injects the first denial
4. repeat allowed + denied
5. receive the allowed response again

Expected: second denial gets injected too.
Actual before this patch: only the allowed response passes, no inject. oops.

Added that sequence to the existing memcached test, so it fails before the fix and passes after.

Tested:

```
go test ./proxylib/... -count=1
```


sayboras:
Thanks for your contribution, however, proxylib is removed in cilium/cilium recently, and will be removed in cilium/proxy in near future.

https://github.com/cilium/cilium/pull/44610
https://github.com/cilium/proxy/issues/1763

PR #1897, opened 2026-05-08
ci: define runner fallback on ci-tests.yaml

mhofstetter:
Currently, ci-tests.yaml fails no runner is defined.

```
Error when evaluating 'runs-on' for job 'tests'. .github/workflows/ci-tests.yaml (Line: 41, Col: 14): Unexpected value ''
```

The reason is that the runner is defined as GH action variable that isn't in the context for workflows that run with `pull_request` (which is the case for `ci-tests.yaml`).

Therefore, this commit adds a fallback runner. Unfortunately this comes with a maintenance burden of periodically updating that value if necessary.

See https://github.com/cilium/proxy/actions/runs/25546244582

### DIFF
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -38,7 +38,7 @@ jobs:
   tests:
     timeout-minutes: 360
     name: Run integration tests on amd64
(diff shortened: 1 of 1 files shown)


sayboras (approved):
:amazing:

PR #1905, opened 2026-05-17
tools: Add adaptive Bazel wrapper

jrajahalme:
NOTE: This bazel wrapper `tools/bazel_adaptive.py` is NOT used by default. Set the `BAZEL` environment variable to use it with `make`, e.g., `BAZEL=tools/bazel_adaptive.py make tests`.

`tools/bazel_adaptive.py` runs Bazel with a concrete integer --jobs value, monitors progress output and memory pressure, stops and resumes bazel build processes as needed under memory pressure, and if that does not relieve memory pressure, restarts with lower or higher parallelism as memory allows. Integer --jobs values are used directly; Bazel-style HOST_CPUS and HOST_RAM values with optional multipliers, such as HOST_CPUS*.5 and HOST_RAM*.0002, are resolved once as the maximum adaptive jobs cap.

The wrapper rewrites bazel output to indicate the number of paused jobs (if any).  "(13 actions, 12 running)" is shown as "(13 actions, 2 paused, 10 running)" when 2 of the running actions have been paused by the wrapper.

This wrapper is only modestly faster than builds with carefully tuned --jobs count and ample available swap space. The main benefit of this wrapper is that it removes the tuning work and automatically restarts bazel when if fails for out-of-memory or job termination.

To allow Bazel to get through memory bottlenecks during the build without unnecessary restarts the build environment should have swap available. This wrapper was tested on a 24 GB VM with 16 GB swap.

Running the wrapper with BAZEL_ADAPTIVE_DISABLE_PTY=1 defined allows more of the diagnostic lines to remain visible even when running in an interactive terminal. Each diagnostic line starts with "[bazel-adaptive/<XXX>s]", where "<XXX>" is the number of seconds since the wrapper started.

Written by Codex through an iterative session where codex has run the included Python tests and has observed the behavior of live runs for runtime behavior.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,6 @@
 
 # clangd compilation database
 /compile_commands.json
+
+# Python caches
+__pycache__
--- a/README.md
+++ b/README.md
@@ -86,6 +86,53 @@ changes refresh it automatically.
 > for the build stages.
 
 
+### Optimizing Memory for Bazel Builds
+
+Sometimes Bazel builds crash due to lack of memory, especially when building
+heavily templated code such as integration tests. The most straightforward
+solution is to lower the number of build jobs via the `--jobs` parameter.
+Makefiles in this repo already scale the number of jobs to the available memory
+when launching Bazel. Depending on the build environment this may leave many or
+even most of the available CPUs underutilized. There are two additional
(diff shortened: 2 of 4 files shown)


jrajahalme:
Added a section in `README.md`.

PR #1909, opened 2026-05-18
build: Simplify bazel command lines

jrajahalme:
Move variables and options to `.bazelrc` files to simplify Bazel command lines. This makes local builds easier and allows for the explicit platform setting to be removed from local builds so that Bazel chooses the appropriate toolchain. For example, instead of:
```
CARGO_BAZEL_REPIN=true bazel test --platforms=//bazel:linux_aarch64 --config=release --extra_toolchains=//bazel/toolchains:all --repo_env=BAZEL_LLVM_PATH=/usr/lib/llvm-18 //tests/...
```
this can be used:
```
bazel test --config=release //tests/...
```
Change clang install to check the installed version and not issuing apt install if not needed. This gets around current upstream clang apt repo breakage with package version requirements.

Note that since `--platforms` option changed Bazel will build from scratch (once).

Remove the repeated bazel cargo repin by bringing the lock file into this repo and adding a new make target `cargo-repin` to update it when needed. This removes about 15 seconds from each bazel invocation.

Fix local build installation.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -6,8 +6,10 @@
 # curl -sSL https://raw.githubusercontent.com/envoyproxy/envoy-wasm/master/.bazelrc > envoy.bazelrc
 import %workspace%/envoy.bazelrc
 
-# Rust crate repinning for repository rules.
-common --repo_env=CARGO_BAZEL_REPIN=true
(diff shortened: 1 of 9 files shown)


nezdolik (commented):
lgtm, thanks!

sayboras (approved):

PR #1917, opened 2026-05-24
envoy: Bump to version v1.38.3

sayboras:


patches/0006-test-integration-Defer-fake-upstream-read-enable-un.patch
```diff
@@ -1,7 +1,19 @@
+From 1d2c987eee776d0a0ac1df6724edf02d9bb4da3a Mon Sep 17 00:00:00 2001
+From: Tam Mach <sayboras@yahoo.com>
+Date: Sat, 23 May 2026 22:46:20 +1000
```
nezdolik: i believe the patch has been merged to upstream, do we still need it here?

sayboras: This fix landed on main (envoyproxy/envoy#45029) after release/v1.38 was cut and wasn't backported, so we still need this patch until we bump to v1.39+.

nezdolik: can we sync the patch with whats been merged to upstream? During review there was an issue uncovered with that patch (wrongly re-enable reads in tests that intentionally keep them disabled): https://github.com/envoyproxy/envoy/pull/45029#discussion_r3226791549

sayboras: good point :100: :heavy_check_mark:

nezdolik (approved):

PR #1919, opened 2026-05-24
bpf_metadata: Override unimplemented Delta xDS in npds_config

jrajahalme:
We do not yet implement Delta xDS so downgrade it to SotW in npds_config. This is needed for upgrade/downgrade support, where an upgraded agent may ask an v1.36 proxy to use Delta xDS. When this is done without this downgrade, Delta SDS will be used for Secrets referenced in the network policy. The Delta mode for SDS then remains stuck for the lifetime of the proxy, even if Cilium Agent later downgrades to a version that does not implement Delta SDS.


sayboras (approved):
✔️

PR #1925, opened 2026-05-28
fix: prevent index-out-of-bounds panics in text memcached parser

immanuwell:
text memcached parser panics on malformed input. any client can trigger it by sending `\r\n` (empty command) or a truncated storage command like `set key\r\n` (missing fields).

root cause: `OnData` tokenizes with `bytes.Fields` but never checks if the result is empty before accessing `tokens[0]`. same story for per-command token counts - storage needs 5 tokens, no guard exists. `replyQueue[0]` is also read without checking if the queue is empty, so an unexpected server reply hits the same fate.

fix: add `len` checks before each unsafe slice access, return `proxylib.ERROR` instead of panicking. added 4 test cases covering the crash paths (empty command, truncated storage/delete/gat).

how to reproduce (before fix):

```go
// in proxylib/libcilium/proxylib_memcached_test.go, add to textTestCases:
CheckOnData(t, 1, false, false, &[][]byte{[]byte("\r\n")}, ...)
// -> panic: runtime error: index out of range [0] with length 0
```

or just run:

```
go test ./proxylib/... -run TestMemcache/text_empty_command_line_returns_error
```

related: #1889

### DIFF
--- a/proxylib/libcilium/proxylib_memcached_test.go
+++ b/proxylib/libcilium/proxylib_memcached_test.go
@@ -571,6 +571,39 @@ var textTestCases = []testCase{
 			}, proxylib.OK, "")
 		},
 	},
+	{
+		"text empty command line returns error",
+		"",
+		func(t *testing.T) {
+			CheckOnData(t, 1, false, false, &[][]byte{[]byte("\r\n")}, []ExpFilterOp{},
+				proxylib.PARSER_ERROR, "")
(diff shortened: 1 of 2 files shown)


sayboras:
Thanks for your contribution, however, proxylib is deprecated and will be removed later. 

https://github.com/cilium/proxy/issues/1763

PR #1927, opened 2026-05-31
gh: Bump clang tidy timeout to 2 hours

jrajahalme:
Clang tidy is very slow, and we have hit the 1 hour timeout on the GH action on large PRs. Bump the timeout to 2 hour so that even larger PRs can get throught the workflow.


sayboras (approved):

sayboras:
Let's see if it's improving, otherwise, we will need to bump the runner box.

PR #1940, opened 2026-06-05
chore: replace absl::optional/nullopt with std equivalents

0xch4z:
Replaces deprecated absl::optional, absl::nullopt with std::optional and std::nullopt.

### DIFF
--- a/cilium/bpf_metadata.cc
+++ b/cilium/bpf_metadata.cc
@@ -8,6 +8,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <memory>
+#include <optional>
 #include <string>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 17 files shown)


sayboras (commented):
While absl::optional could be deprecated, we might need to wait/propose changes in upstream first

https://github.com/envoyproxy/envoy/blob/9297c23ebfdc5ca2171873261d01a7d65ff75b32/tools/code_format/check_format.py#L691-L692

nezdolik:
i will check if anyone in upstream is working on this already.

nezdolik:
Upstream should be unblocked: https://github.com/envoyproxy/envoy/pull/45607 (will be part of 1.39)

PR #1944, opened 2026-06-08
bpf_metadata: select ingress source IP by destination family

elohmeier:
North/south L7 LB metadata currently selects the configured local ingress source address using the incoming source address family. When a load balancer forwards an IPv6 client address via PROXY protocol to an IPv4-only node/backend path, the source address is IPv6 but the destination connection is IPv4. In that case the metadata filter looks for an IPv6 ingress source address even though the outbound connection needs the configured IPv4 source address, and metadata extraction fails.

This changes the source-address selection to use the destination address family instead. That matches the address family used by the outbound connection while still preserving the remote source address from PROXY protocol for downstream metadata.

A regression test covers an IPv6 remote source with only an IPv4 local ingress source address configured.

Validation:
- `/home/linuxbrew/.linuxbrew/opt/llvm/bin/clang-format --dry-run -Werror cilium/bpf_metadata.cc tests/bpf_metadata_config_test.cc`
- `git diff --check`
- `bazel test --copt=-Wno-nullability-completeness //tests:bpf_metadata_config_test` could not complete in my local ostree environment after rebasing because the repo toolchain expects `/usr/bin/clang`, which is not present and `/usr` is read-only. The change was also built into a Cilium Envoy image and smoke-tested behind a Cilium Gateway with PROXY protocol: IPv4 and IPv6 requests both succeeded, and the backend received the expected `x-forwarded-for` / `x-envoy-external-address` client IPs.

### DIFF
--- a/cilium/bpf_metadata.cc
+++ b/cilium/bpf_metadata.cc
@@ -462,16 +462,18 @@ Config::extractSocketMetadata(Network::ConnectionSocket& socket) {
     // North/south L7 LB, assume the source security identity of the configured source addresses,
     // if any and policy for this identity exists.
 
-    // Pick the local ingress source address of the same family as the incoming connection
-    const Network::Address::Ip* ingress_ip = selectIpVersion(sip->version(), source_addresses);
+    // Pick the local ingress source address of the same family as the destination connection.
+    // The incoming source family may differ when an IPv6 client reaches an IPv4-only load-balancer
+    // path via PROXY protocol.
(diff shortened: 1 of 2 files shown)


jrajahalme (approved):
Thanks for the fix and the regression test. I'll look into suppressing the info log on the address type mismatch as it is expected in this situation.

elohmeier:
This is now included when using the Cilium 1.20.2 chart.

PR #1950, opened 2026-06-13
docs: refresh README version compatibility matrix

immanuwell:
README version matrix is kinda stale.

`tools/update_version_matrix.sh --dry-run` prints newer supported Cilium patch releases, but `README.md` still stops earlier. `(main)` is also behind and still says `v1.36.x`.

This just refreshes the generated table in `README.md`. no runtime code, no build logic, just docs sync.

How to repro:
1. check out `main`
2. run `SUPPORTED_MINOR_VERSIONS=3 ./tools/update_version_matrix.sh --dry-run`
3. compare the printed table with `README.md`
4. you will see missing rows like `v1.19.4`, `v1.18.10`, `v1.17.16`, and `(main)` still on `v1.36.x`

Double checked this against current `cilium/cilium` release tags and `images/cilium/Dockerfile`, so the table lines up now. pretty small one, but useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,15 @@ Note: The below table is updated by script `tools/update_version_matrix.sh`
 
 | Cilium Version | Envoy version |
 |----------------|---------------|
-| (main)         | v1.36.x       |
+| (main)         | v1.37.x       |
+| v1.19.4        | v1.36.6       |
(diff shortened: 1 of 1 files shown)


sayboras (approved):
Thanks a lot for your help 💯

PR #1964, opened 2026-07-01
fix: handle missing generic L7 metadata entries

immanuwell:
generic `l7_proto` enforcement in `cilium.network` assumes `dynamicMetadata().filter_metadata()[l7proto]` is always there.
when that namespace is missing, `onData()` hits `.at(l7proto_)` and throws. not great.

this treats the missing namespace as empty metadata and keeps the policy path running. also adds a small regression test.

repro
1. configure a generic `l7_proto` policy with `metadata_rule`
2. hit `cilium.network::onData()` before any filter writes `dynamicMetadata().filter_metadata()[l7proto]`
3. current code throws on `.at(l7proto_)`

note: i couldnt run `bazel test //tests:network_filter_test` here, local bazel gets stuck earlier on a `rules_foreign_cc` toolchain bootstrap failure.

### DIFF
--- a/tests/network_filter_test.cc
+++ b/tests/network_filter_test.cc
@@ -1,0 +1,96 @@
+#include <cstdint>
+#include <memory>
+#include <string>
+#include <utility>
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include "cilium/accesslog.h"
+#include "cilium/api/network_filter.pb.h"
+#include "cilium/filter_state_cilium_policy.h"
+#include "cilium/network_filter.h"
+#include "cilium/network_policy.h"
+
+#include "envoy/network/address.h"
+#include "envoy/network/filter.h"
+#include "envoy/stream_info/filter_state.h"
+#include "envoy/stream_info/stream_info.h"
+#include "envoy/upstream/host_description.h"
+
+#include "source/common/buffer/buffer_impl.h"
+
+#include "test/mocks/network/mocks.h"
+#include "test/mocks/server/listener_factory_context.h"
+
+namespace Envoy {
+namespace Filter {
+namespace CiliumL3 {
+
+class NetworkFilterTestPeer {
+public:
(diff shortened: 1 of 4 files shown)


tests/network_filter_test.cc
```diff
@@ -0,0 +1,54 @@
+#include <gtest/gtest.h>
+
+#define private public
```
nezdolik: why is this needed? if you need access to private member fields to validate in test, consider declaring friend peer class, e.g. https://github.com/cilium/proxy/blob/main/cilium/grpc_subscription.h#L67 and then define accessor methods in test file (network_filter_test.cc) like so https://github.com/cilium/proxy/blob/7cffc778c923f68a77954a53b1a98d6b5353f004/tests/cilium_test_peer.h#L17

nezdolik (approved):

nezdolik:
@immanuwell please fix ci checks

tests/network_filter_test.cc
immanuwell: updated to use the peer-class approach you suggested

immanuwell:
@nezdolik fixed all CI checks, and switched to peer-class approach you suggested

nezdolik:
@immanuwell more ci checks need fixing

immanuwell:
@nezdolik I am sorry but now after latest push all CI checks should be green
can you trigger it please?

immanuwell:
remaining CI failure occurs before PR code runs: pull_request_target blocks checking out fork code for safety

nezdolik:
@immanuwell raised a fix here https://github.com/cilium/proxy/pull/1975 for `pull_request_target`. Once it gets merged you will need to rebase on top of latest main, that should fix asan and "ci build & push"  ci workflows. The format ci failure is legit though and needs to be addressed by you.

immanuwell:
@nezdolik rebased onto that #1975

PR #1969, opened 2026-07-06
ci: add ASAN integration test job

nezdolik:
Running the existing integration tests under AddressSanitizer and UndefinedBehaviorSanitizer gives us a targeted signal for use-after-free, buffer overflows, invalid lifetime assumptions, and undefined behavior in exercised proxy code paths.

The ASAN job reuses the Dockerfile.tests flow with reduced parallelism and longer timeouts to keep sanitizer output readable. The job passes the sanitizer Bazel configs explicitly so the check stays scoped to the Cilium proxy CI path.


nezdolik:
to make sure ci changes are picked up raised a PR not from the form but from the upstream branch: https://github.com/cilium/proxy/pull/1970

PR #1970, opened 2026-07-06
ci: add ASAN integration test job

nezdolik:
Running the existing integration tests under AddressSanitizer and UndefinedBehaviorSanitizer gives us a targeted signal for use-after-free, buffer overflows, invalid lifetime assumptions, and undefined behavior in exercised proxy code paths.

The ASAN job reuses the Dockerfile.tests flow with reduced parallelism and longer timeouts to keep sanitizer output readable. The job passes the sanitizer Bazel configs explicitly so the check stays scoped to the Cilium proxy CI path.

### DIFF
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -83,3 +83,53 @@ jobs:
             BAZEL_TEST_OPTS=--jobs=HOST_RAM*.0002 --test_timeout=300 --local_test_jobs=4 --flaky_test_attempts=3
           cache-from: type=local,src=/tmp/buildx-cache
           push: false
+
+  asan-tests:
+    timeout-minutes: 360
+    name: Run ASAN integration tests on amd64
+    runs-on: ${{ vars.PROXY_BUILD_GITHUB_RUNNER || 'oracle-vm-32cpu-128gb-x86-64' }}
+    steps:
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@d7f5e7f509e45cec5c76c4d5afdd7de93d0b3df5 # v4.1.0
+
+      - name: Cache Docker layers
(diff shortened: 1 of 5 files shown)


mhofstetter (commented):
1. It seems as the added test fails on the PR.
2. Please remove the merge commit and use rebase

nezdolik:
right, asan check spotted some problematic code parts, but symbolizer is not giving human readable stack traces. I will try to fix symbolizer as well as part of this PR, then failing asan checks need to be fixed.

nezdolik:
symbolizer is fixed now, now will be addressing asan failures one by one (asan stops after the first error, so more may appear once current one is fixed). Current failure:
heap-use-after-free in AllowedPathRegexDeleteHeader/0:
[network_policy.cc:460](https://github.com/cilium/proxy/blob/129aa95711e87e8a253a9326b33d47aa86b7c890/cilium/network_policy.cc#L460) obtains a view into the request header.
[network_policy.cc:524](https://github.com/cilium/proxy/blob/129aa95711e87e8a253a9326b33d47aa86b7c890/cilium/network_policy.cc#L524) removes the header, freeing its storage.
[network_policy.cc:525](https://github.com/cilium/proxy/blob/129aa95711e87e8a253a9326b33d47aa86b7c890/cilium/network_policy.cc#L525) passes the now-dangling view to logRejected().
[accesslog.cc:277](https://github.com/cilium/proxy/blob/129aa95711e87e8a253a9326b33d47aa86b7c890/cilium/accesslog.cc#L277) reads that freed memory while copying the value into protobuf.
The trigger is [cilium_http_integration_test.cc:775](https://github.com/cilium/proxy/blob/129aa95711e87e8a253a9326b33d47aa86b7c890/tests/cilium_http_integration_test.cc#L775).

nezdolik:
@mhofstetter the CI is now green.

mhofstetter (approved):

PR #1978, opened 2026-07-17
test: add temporary MSAN failure to test ci

nezdolik:
Deliberately read uninitialized heap memory in an MSAN-only test. This should make msan ci check to fail.


nezdolik:
failed as expected

PR #2002, opened 2026-08-17
patches: convert gRPC mux callback patch format

nezdolik:
Regenerate the patch with git format-patch metadata.


sayboras (approved):
Thanks 💯

PR #2006, opened 2026-08-20
policy: Deduplicate SecretWatchers and TLS contexts

jrajahalme:
Network policies may contain hundreds or thousands of HeaderMatch or TLS rules referring to a small set of Secrets. Each reference previously constructed a separate SecretWatcher or Envoy TLS ContextConfigImpl, including its SDS provider callbacks and other context state.

Add separate weak caches for SecretWatchers and upstream and downstream TLS contexts, keyed by the SDS name or Cilium TLSContext protobuf, respectively. Reuse a live context when an equivalent policy configuration is encountered.

Reset the caches together with the policy maps after an NPDS stream restart, preserving the existing agent-restart behavior. Prune expired cache entries asynchronously after policy updates and worker quiescence, keeping pruning off the policy-update hot path.

Benchmark an NPDS update containing 1,000 policies that reference the same SDS-derived CA validation context. The CA is delivered through a real dynamic SDS provider rather than embedded in the policy TLS contexts.

Optimized benchmark results (3 repetitions):
```
                                CPU time     Retained heap     Heap/policy
  TLS context cache enabled     1.47 ms      1.581 MiB         1.619 KiB
  TLS context cache disabled    3.95 ms      4.990 MiB         5.109 KiB
```
This reduces CPU time by approximately 63% and retained heap by 68%, saving about 3.41 MiB per 1,000-policy update and increasing throughput by 2.68x.

The benchmark uses Envoy's small, single-certificate test CA. Production SDS resources commonly contain larger CA bundles comparable to those shipped with operating systems, so the benchmark likely understates the memory impact of retaining duplicate certificate-related context state.

Benchmark results for deduplication of SecretWatchers on HeaderMatches are more modest (duplicating header sized secrets carries less overhead), but are still positive around 20-25% for CPU and memory.

### DIFF
--- a/cilium/network_policy.cc
+++ b/cilium/network_policy.cc
@@ -299,7 +299,8 @@ class NetworkPolicyMapImpl : public ManagedGrpcSubscription {
 public:
   friend class PortNetworkPolicyRule;
   NetworkPolicyMapImpl(Server::Configuration::FactoryContext& context,
-                       const envoy::config::core::v3::ConfigSource& config_source, bool subscribe);
+                       const envoy::config::core::v3::ConfigSource& config_source, bool subscribe,
+                       bool policy_secret_cache_enabled);
   ~NetworkPolicyMapImpl() override;
 
   // Config::SubscriptionCallbacks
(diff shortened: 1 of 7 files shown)


jrajahalme:
added a missing include for tidy

jrajahalme:
restarted integration test due to a flake fixed by https://github.com/cilium/proxy/pull/2007

cilium/network_policy.cc
```diff
@@ -2107,9 +2124,18 @@ void NetworkPolicyMapImpl::scheduleDeferredDeletion(const PolicyMapSnapshot* old
   if (old_policy_map == nullptr) {
     return;
   }
-  runAfterAllThreads([old_policy_map]() {
+  const std::weak_ptr<NetworkPolicyMapImpl> weak_this =
```
nezdolik: nit: to avoid those casts `NetworkPolicyMapImpl` could inherit from `std::enable_shared_from_this`:
```
class NetworkPolicyMapImpl : public std::enable_shared_from_this<NetworkPolicyMapImpl>
```
and then you can simply invoke `weak_from_this()` here.

example of how it is done in geoip filter:
https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/geoip/geoip_filter.h#L48
https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/geoip/geoip_filter.cc#L76

cilium/secret_watcher.cc
```diff
@@ -240,5 +245,132 @@ UpstreamTLSContext::UpstreamTLSContext(
...
+  auto it = cache.find(key);
+  if (it != cache.end()) {
+    if (auto cached = it->second.lock()) {
+      return cached;
+    }
+    cache.erase(it);
+  }
+
+  auto value = factory();
+  cache.emplace(key, value);
```
nezdolik, line 278: what would happen when entire npds update is rejected? will those already inserted entries be rolled back or scheduled for prunning?

cilium/network_policy.cc
jrajahalme: The base class already inherits from `std::enable_shared_from_this`, and having the same class inherit from two different specializations of `std::enable_shared_from_this` could be ambiguous.

I added a helpers  `sharedFromThis()` and `weakFromThis()` that hides the cast, already had one of these in a feature branch I am working on.

cilium/secret_watcher.cc
jrajahalme: Not rolled back, but pruned later. Immediate pruning should not be necessary.

nezdolik (approved):

PR #2007, opened 2026-08-20
cilium: Separate NPHDS subscription and host map stats

jrajahalme:
Use the cilium.nphds scope for the generic gRPC subscription stats while retaining cilium.hostmap for stats tracking successfully applied host map updates.

Previously both sets of stats shared the cilium.hostmap scope, causing update_success to be incremented twice for each accepted NPHDS response. Tests waiting on this counter could consequently proceed before the corresponding host map update had been installed, resulting in a timing-dependent assertion failure under ASAN.

This follows the existing policy map convention, where NPDS subscription stats and successfully applied policy update stats use separate scopes.

Since NPHDS is not used in production this only showed up as CI test flakes like this:
```
#18 221.5 tests/bpf_metadata_integration_test.cc:648: Failure
#18 221.5 Expected equality of these values:
#18 221.5   resolveHostPolicyId("10.1.1.1")
#18 221.5     Which is: 111
#18 221.5   Cilium::ID::UNKNOWN
#18 221.5     Which is: 0
```

### DIFF
--- a/cilium/host_map.cc
+++ b/cilium/host_map.cc
@@ -205,10 +205,11 @@ PolicyHostMap::PolicyHostMap(Server::Configuration::CommonFactoryContext& contex
                              bool subscribe)
     : ManagedGrpcSubscription(
           NetworkPolicyHostsTypeUrl, []() { return std::make_shared<Cilium::PolicyHostDecoder>(); },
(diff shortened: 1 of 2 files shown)


nezdolik (approved):
looks like ci needs fixing

PR #2012, opened 2026-08-25
Fix websocket test flakes and complete TCP half-closed support

jrajahalme:
This PR fixes two WebSocket tunnel shutdown issues exposed by sanitizer CI, completes tunneled TCP half-close support, and tightens frame decoding to RFC 6455.

The first three commits form a self-contained backport set. They give flush tests enough buffer headroom for WebSocket framing, make test shutdowns use an orderly CLOSE exchange, and fix the production race between final frames and transport shutdown. CLOSE remains a directional FIN for the tunneled TCP stream, but the client codec no longer attaches transport end-stream to its final frames. Only the server codec explicitly ends the transport after both tunnel directions have ended, allowing queued frames to be flushed before shutdown.

The remaining commits complete bidirectional half-close semantics. Receiving CLOSE ends the corresponding tunneled TCP direction exactly once while reverse-direction data and control traffic remain active. A CLOSE response is delayed until reverse data has drained, and it preserves the received status and reason payload. PING and PONG continue to work during the half-closed interval.

The PR carries two Envoy patches that lets write filters consume data after decoded end stream before Envoy rejects any unconsumed write, and delays the close of an upstream socket on a TCPProxy connection when the downstream closes. The first patch lets the WebSocket filter handle control traffic such as PING and generate PONG after the tunneled TCP direction has ended without permitting application data to be written after FIN. It is the write-side counterpart to [envoyproxy/envoy#38189](https://github.com/envoyproxy/envoy/pull/38189). The second patch allows any buffered final data to be encoded to the WebSocket tunnel, and WebSocket CLOSE handshake to finish gracefully after downstream TCP connection closes.

The tests now use correctly masked client frames and cover both half-close orderings, final data, flushing, control traffic, CLOSE payload propagation, and aborts. Decoder hardening rejects incorrect masking, fragmented or oversized control frames, malformed CLOSE payloads, invalid UTF-8 reasons, and transport FIN without CLOSE.

### DIFF
--- a/cilium/websocket.cc
+++ b/cilium/websocket.cc
@@ -1,5 +1,6 @@
 #include "cilium/websocket.h"
 
+#include <chrono>
 #include <cstdint>
 #include <memory>
 #include <string>
@@ -19,6 +20,7 @@
 #include "source/common/protobuf/protobuf.h"
 #include "source/common/protobuf/utility.h"
 #include "source/common/stream_info/bool_accessor_impl.h"
+#include "source/common/stream_info/uint64_accessor_impl.h"
 #include "source/common/tcp_proxy/tcp_proxy.h"
 
 #include "absl/status/statusor.h"
@@ -33,6 +35,8 @@ namespace Cilium {
 namespace WebSocket {
 
 namespace {
+
+constexpr std::chrono::milliseconds WebSocketTransportCloseTimeout{1000};
 
 Http::RegisterCustomInlineHeader<Http::CustomInlineHeaderRegistry::Type::RequestHeaders>
     origin_handle(Http::CustomHeaders::get().Origin);
@@ -128,6 +132,16 @@ void Instance::initializeReadFilterCallbacks(Network::ReadFilterCallbacks& callb
   callbacks_->connection().streamInfo().filterState()->setData(
       TcpProxy::ReceiveBeforeConnectKey, std::make_unique<StreamInfo::BoolAccessorImpl>(true),
       StreamInfo::FilterState::StateType::Mutable, StreamInfo::FilterState::LifeSpan::Connection);
+
+  // After both directions of a client-side WebSocket tunnel have ended, TcpProxy must flush the
+  // final data and CLOSE frame and wait for the peer's transport FIN. Closing immediately can
+  // generate an RST if a peer control frame is still unread, losing the just-flushed frames.
+  if (config_->client_) {
+    callbacks_->connection().streamInfo().filterState()->setData(
+        TcpProxy::UpstreamFlushWaitTimeoutMs,
+        std::make_unique<StreamInfo::UInt64AccessorImpl>(WebSocketTransportCloseTimeout.count()),
+        StreamInfo::FilterState::StateType::Mutable, StreamInfo::FilterState::LifeSpan::Connection);
+  }
 }
 
 Network::FilterStatus Instance::onNewConnection() {
--- a/cilium/websocket_codec.cc
+++ b/cilium/websocket_codec.cc
@@ -373,22 +373,45 @@ void Codec::handshake() {
 void Codec::encode(Buffer::Instance& data, bool end_stream) {
   ENVOY_LOG(debug, "websocket: encode {} bytes, end_stream: {}", data.length(), end_stream);
 
-  encoder_.encode(data, end_stream, OPCODE_BIN);
+  // RFC 6455 section 5.5.1 forbids sending data frames after a CLOSE frame.
+  if (close_sent_) {
+    const auto len = data.length();
+    if (len > 0) {
+      ENVOY_LOG(debug, "websocket encoder: data received after CLOSE: {} bytes", len);
+      data.drain(len);
+    }
+    return;
+  }
+
+  encoder_.encode(data, OPCODE_BIN);
+
+  if (end_stream) {
+    // CLOSE represents the FIN for this direction of the tunneled TCP connection. If the peer
+    // half-closed first, echo its validated CLOSE payload only after all reverse-direction data
+    // has been encoded. This deliberately delays the CLOSE response to preserve TCP half-close
+    // semantics when the WebSocket infrastructure permits it.
+    Buffer::OwnedImpl close_payload;
+    close_payload.add(decoder_.close_payload_);
+    encoder_.encode(close_payload, OPCODE_CLOSE);
+    close_sent_ = true;
+
+    // Only end stream if CLOSE has been received already and if we are the server, as additional
+    // control frames may still be sent, and RFC 6455 section 7.1.1 recommends that the server close
+    // the underlying transport after both sides have sent CLOSE. Keep the client transport open so
+    // that its final frames are flushed before the server closes the connection.
+    end_stream = !config()->client_ && decoder_.close_received_;
+  }
 
   // Only forward data if handshake has completed
   if (accepted_) {
-    // Reset idle timer on data
+    // Reset idle timer on any frame content
     if (encoder_.hasData()) {
       resetPingTimer();
     }
-    // A WebSocket CLOSE represents the FIN for this direction of the tunneled TCP connection.
-    // RFC 6455 section 7.1.1 recommends that the server close the underlying transport after both
-    // sides have sent CLOSE. Keep the client transport open so that its final frames are flushed
-    // before the server closes the connection.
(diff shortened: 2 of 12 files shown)


patches/0010-tcp_proxy-Allow-filters-to-wait-for-upstream-close.patch
```diff
@@ -0,0 +1,260 @@
...
+    * @param event supplies the event which occurred.
+    * @param details supplies the details of the event, e.g. the local close reason.
++   * @param flush_wait_timeout supplies how long a TCP upstream should wait for its peer to close
++   *        after pending writes have been flushed. Zero preserves the normal immediate close.
+    * @return the underlying ConnectionData if the event is not "Connected" and draining
+              is supported for this upstream.
+    */
+   virtual Tcp::ConnectionPool::ConnectionData*
+-  onDownstreamEvent(Network::ConnectionEvent event, absl::string_view details = "") PURE;
++  onDownstreamEvent(Network::ConnectionEvent event, absl::string_view details = "",
```
nezdolik, line 45: by changing this signature there is possibility of clashes for extensions that override this method? or we dont use any of such extensions?

cilium/websocket_codec.cc
```diff
@@ -654,118 +685,120 @@ void Codec::decode(Buffer::Instance& data, bool end_stream) {
...
+  // WebSocket clients must mask all frames, including control frames.
+  if (config->client_) {
+    frame_header[1] |= MASK_MASK;
+    const uint32_t mask_word = config->random_.random();
+    uint8_t mask[4];
+    for (size_t i = 0; i < sizeof(mask); ++i) {
+      mask[i] = static_cast<uint8_t>(mask_word >> (i * 8));
+      frame_header[frame_header_length++] = mask[i];
+    }
+    uint8_t* buf = reinterpret_cast<uint8_t*>(data.linearize(payload_len));
```
nezdolik, line 795: `linearize()` seems to take `uint32_t`, so there is a risk of narrowing conversions on 64-bit Envoy builds. The problem was not introduced by this pr but just an observation

cilium/websocket.cc
```diff
@@ -128,6 +132,16 @@ void Instance::initializeReadFilterCallbacks(Network::ReadFilterCallbacks& callb
...
       StreamInfo::FilterState::StateType::Mutable, StreamInfo::FilterState::LifeSpan::Connection);
+
+  // After both directions of a client-side WebSocket tunnel have ended, TcpProxy must flush the
+  // final data and CLOSE frame and wait for the peer's transport FIN. Closing immediately can
+  // generate an RST if a peer control frame is still unread, losing the just-flushed frames.
+  if (config_->client_) {
+    callbacks_->connection().streamInfo().filterState()->setData(
+        TcpProxy::UpstreamFlushWaitTimeoutMs,
+        std::make_unique<StreamInfo::UInt64AccessorImpl>(WebSocketTransportCloseTimeout.count()),
+        StreamInfo::FilterState::StateType::Mutable, StreamInfo::FilterState::LifeSpan::Connection);
```
nezdolik, line 143: does the value need to be mutable?

tests/cilium_websocket_codec_integration_test.cc
```diff
@@ -7,6 +7,9 @@
 #include <cstdint>
 #include <string>
 
+#include "envoy/common/platform.h"
```
nezdolik, line 10: nit: can you add explicit dependencies for this includes into test BUILD target?

tests/cilium_websocket_codec_integration_test.cc
```diff
@@ -134,9 +137,22 @@ TEST_P(CiliumWebSocketIntegrationTest, CiliumWebSocketUpstreamWritesFirst) {
   ASSERT_TRUE(fake_upstream_connection->waitForData(5, &received));
   ASSERT_EQ(received, "hello");
 
-  ASSERT_TRUE(fake_upstream_connection->write("", true));
+  // A FIN in one direction must not prevent data from flowing in the reverse direction. The first
+  // CLOSE crosses both WebSocket codecs and becomes a half-close at the TCP client.
+  ASSERT_TRUE(fake_upstream_connection->write("upstream final", true));
+  tcp_client->waitForData("helloupstream final");
```
nezdolik, line 143: here and elsewhere in the tests it would more reliable to also assert the exact received data like so:
```
ASSERT_EQ("helloupstream final", tcp_client->data());
```

patches/0010-tcp_proxy-Allow-filters-to-wait-for-upstream-close.patch
jrajahalme: This can actually be done without changing the API visible to the extensions at all, changed!

cilium/websocket_codec.cc
jrajahalme: good call; while it is very unlikely that we'd get more than 4GB data from a TCP socket at once, this is still a correctness issue. Changed to mask in place in the buffer slices without linearizing.

cilium/websocket.cc
jrajahalme: No, `ReadOnly` is a better fit, thanks!

tests/cilium_websocket_codec_integration_test.cc
jrajahalme: We haven't done that for any test targets, the new includes are already covered by the existing bazel deps. Doing this would need a linter to enforce this, and IMO that is not worth the effort/churn, at least for now.

tests/cilium_websocket_codec_integration_test.cc
jrajahalme: It might be better to do this in a separate PR, as it will touch not just the websocket tests?

jrajahalme: It looks like maybe EXPECT_TCP_RESPONSE is the right replacement here, as it combines the wait and assert without having to repeat the string?

jrajahalme: turns out only the TCP integrations tests use `IntegrationTcpClient::waitForData` in addition to the websocket integration tests, so might do this in this PR after all...

jrajahalme: Added this as a separate commit, and doing this actually exposed one malformed test case that accidentally passed due to waitForData returning when the TCP stream ends.

jrajahalme:
@nezdolik Thanks for the thorough review, updated :-)

jrajahalme:
final push (hopefully) addressing clang tidy false positive in EXPECT_TCP_RESPONSE.

patches/0010-tcp_proxy-Allow-filters-to-wait-for-upstream-close.patch
```diff
@@ -0,0 +1,316 @@
...
+-      upstream_host_(upstream_host), config_(config) {
++      upstream_host_(upstream_host), config_(config), discard_upstream_data_(discard_upstream_data) {
+   ENVOY_CONN_LOG(trace, "draining the upstream connection", upstream_conn_data_->connection());
+   config_->stats().upstream_flush_total_.inc();
+   config_->stats().upstream_flush_active_.inc();
+@@ -1551,6 +1564,10 @@ void Drainer::onEvent(Network::ConnectionEvent event) {
+ }
+ 
+ void Drainer::onData(Buffer::Instance& data, bool) {
++  if (discard_upstream_data_) {
```
nezdolik: could you add a test where downstream closes, upstream connection gets half closed and enters `FlushWriteAndDelay` and then upstream sends data folowed by FIN? Did not see where read/close events remain enabled in this state, suspecting this code path cannot be reached after delayed close begins.

jrajahalme: will add a test for this. We currently have separate tests for the different parts of this, but not a full integration test yet.

jrajahalme: Added this test (`CiliumWebSocketDownstreamCloseReceivesFinalData`) and another commit that makes the `CiliumWebSocketIntegrationTest` more realistic by splitting it into two TCP proxy listeners, one for the WebSocket client and the other for the WebSocket server with a real socket between them.

nezdolik (approved):

PR #2046, opened 2026-09-18
healthcheck: add new cilium-envoy-healthcheck binary

0xch4z:
Adds a new cilium-envoy-healthcheck binary that takes the path of the envoy admin socket and exits with status 1 if the /ready endpoint returns a non-OK status code.

This can be used in place of the cilium configured /healthz readiness probe and avoid using the host network. This would be a prerequisite to for example effectively use surge pods for envoy.


nezdolik:
could you check failing ci?

nezdolik (approved):

0xch4z:
The check format workflow got a 500 from github pulling down rules_proto_grpc. https://github.com/cilium/proxy/actions/runs/35356114686/job/105635761399?pr=2046 :headstone: 

I opened [another PR](https://github.com/cilium/proxy/pull/2047) to retry those

